PR #32, opened 2019-10-11
Changes the behavior to always include the default strategies

### LINKED ISSUE #10: Default strategies not available when supplying custom strategies
I would have expected the custom strategies to be added to the default ones. Unfortunately the default ones are skipped.

If this is a deliberate design decision, please consider exposing a public property to access the default strategies, so one can choose to pass these in as part of the custom strategies.

Excolo:
fixes #10

- Changes the default behavior of the `DefaultUnleash` constructor to always include the default strategies, regardless of whether any custom strategies were provided. 
- Provides a constructor overload to allow overriding the default strategies. _Beware_, that overriding this without any custom strategies will yield an empty strategy map. I'm sure whether this is acceptable. 
- Fixes a minor typo.

**This implementation will be a breaking change for anyone currently passing custom strategies, expecting only their custom strategies to apply.**

### DIFF
--- a/src/Unleash/DefaultUnleash.cs
+++ b/src/Unleash/DefaultUnleash.cs
@@ -180,7 +180,7 @@ namespace Unleash
         {
             if (overrideDefaultStrategies)
             {
-                return strategies ?? Array.Empty<IStrategy>();
+                return strategies ?? new IStrategy[0];
             }
             else
             {
--- a/src/Unleash/DefaultUnleash.cs
+++ b/src/Unleash/DefaultUnleash.cs
@@ -15,7 +15,7 @@ namespace Unleash
 
         private static readonly UnknownStrategy UnknownStrategy = new UnknownStrategy();
 
-        private static readonly IStrategy[] DefaultStragegies = {
+        private static readonly IStrategy[] DefaultStrategies = {
             new DefaultStrategy(),
             new UserWithIdStrategy(),
             new GradualRolloutUserIdStrategy(),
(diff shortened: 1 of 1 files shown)


src/Unleash/DefaultUnleash.cs
```diff
@@ -30,22 +30,34 @@ public class DefaultUnleash : IUnleash
...
+            : this(settings, overrideDefaultStrategies: false, strategies)
+        { }
+
+        ///// <summary>
+        ///// Initializes a new instance of Unleash client.
+        ///// </summary>
+        ///// <param name="config">Unleash settings</param>
+        ///// <param name="overrideDefaultStrategies">When true, it overrides the default strategies.</param>
+        ///// <param name="strategies">Custom strategies.</param>
+        public DefaultUnleash(UnleashSettings settings, bool overrideDefaultStrategies, params IStrategy[] strategies)
```
ivarconr: I'm not sure if I understand the use case for the `overrideDefaultStrategies`? 

Other client SDKs has implemented this in the following way:
The client has a list of strategies it understand. If the user sends in a custom implementation with the same name as an default strategy the custom strategy will be used.

Excolo: I mostly added it so that people experiencing breaking changes, could fix it by changing their code to:
```C#
var unleash = new DefaultUnleash(settings, true, myStrategy, myOtherStrategy);
```
thus having it behave as it did before. Basically a means to make the breaking changes more acceptable - although, I'm not sure how big of a deal these changes really are.

I do like your suggestion.

src/Unleash/DefaultUnleash.cs
```diff
@@ -30,22 +30,34 @@ public class DefaultUnleash : IUnleash
         private readonly UnleashServices services;
 
         ///// <summary>
-        ///// Initializes a new instance of Unleash client. 
+        ///// Initializes a new instance of Unleash client with a set of default strategies. 
         ///// </summary>
         ///// <param name="config">Unleash settings</param>
-        ///// <param name="strategies">Available strategies. When none defined, all default strategies will be added.</param>
+        ///// <param name="strategies">Additional custom strategies.</param>
         public DefaultUnleash(UnleashSettings settings, params IStrategy[] strategies)
```
silviogreuel: Great, this will keep backwards compatibility 👌

PR #41, opened 2019-10-29
Hotfix/fixing socket leak

amj1985:
# Description

DefaultHttpClientFactory is opening a new socket for each request and its throwing a socket exception in high availability API

## Fixes

https://github.com/Unleash/unleash-client-core/issues/40

## Type of change

# How Has This Been Tested?

Created a new test file **UnleashHttpClientFactory_RegisterHttpClient_Tests** that just validates that when we call **Create** passing an existing uri the factory will return the same instance(using hashcode) and when we call **Create** passing a new uri it will create a new instance.

# Checklist:

### DIFF
--- a/src/Unleash/DefaultHttpClientFactory.cs
+++ b/src/Unleash/DefaultHttpClientFactory.cs
@@ -36,7 +36,7 @@ namespace Unleash
 
             return _httpClientCache.GetOrAdd(key, k =>
             {
-                var client = new HttpClient()
+                var client = new HttpClient
                 {
                     BaseAddress = unleashApiUri,
                     Timeout = Timeout
--- a/src/Unleash/DefaultHttpClientFactory.cs
+++ b/src/Unleash/DefaultHttpClientFactory.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Net;
 using System.Net.Http;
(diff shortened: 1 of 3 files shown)


tests/Unleash.Tests/Communication/UnleashHttpClientFactory_RegisterHttpClient_Tests.cs
```diff
@@ -0,0 +1,40 @@
+﻿using NUnit.Framework;
+using System;
+
+namespace Unleash.Tests.Communication
+{
+    public class UnleashHttpClientFactory_RegisterHttpClient_Tests
+    {
+        private Uri _apiUri { get; } = new Uri("http://unleash.herokuapp.com/");
```
tetexxr: You can use fields instead of properties, it's more simple and more readable.
Apart form that, the .NET properties name convention is not the same as fields.

src/Unleash/DefaultHttpClientFactory.cs
```diff
@@ -22,27 +23,32 @@ public class DefaultHttpClientFactory : IHttpClientFactory
...
 
-            var httpClient = NewHttpClient();
-            httpClient.BaseAddress = unleashApiUri;
-            httpClient.Timeout = Timeout;
-
-            ConfigureHttpClient(httpClient);
-            ConfigureDefaultRequestHeaders(httpClient.DefaultRequestHeaders);
+            return _httpClientCache.GetOrAdd(key, k =>
+            {
+                var client = new HttpClient()
```
tetexxr: You can instance the HttpClient without parenthesis.

tests/Unleash.Tests/Communication/UnleashHttpClientFactory_RegisterHttpClient_Tests.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        private Uri _newApiUri { get; } = new Uri("http://unleash2.herokuapp2.com/");
+
+        private DefaultHttpClientFactory _httpClientFactory { get; } = new DefaultHttpClientFactory();
+
+        [Test]
+        public void HttpClientFactory_Should_Create_Single_Instance_By_Dns_Success()
+        {
+
```
tetexxr: Extra new-lines at the beginning of those methods can be removed.

Issue #11, opened 2019-10-02
Implement client-specification tests


ivarconr:
There exists a test suite which validates correct SDK behaviour. Would be nice if the .net client also could implement these integration tests. 

https://github.com/Unleash/client-specification

PR #45, opened 2020-01-15
Align Variants implementation with Java client

gnab:
Prerequisite for passing client specification tests (#11).


ivarconr:
:fire:

PR #53, opened 2020-02-12
[WIP] added an unleash client factory 

### LINKED ISSUE #50: As a user, I would like to be able to fetch feature toggles synchronously during startup
The Java client for Unleash (https://github.com/Unleash/unleash-client-java) has a flag in UnleashSettings called synchronousFetchOnInitialisation.
Currently, even though feature toggles can be fetched on startup (and are by default), it is not done synchronously.

Expected behavior is to have feature flags loaded immediately after instantiating DefaultUnleash. Observed behavior is that feature flags are loaded in the background, and available some time later. As far as I can tell, use of feature flags on startup requires a short sleep after calling DefaultUnleash, which is unpleasant and racecondition-y.

AndreasChristianson:
# Description
**Work in progress**. I welcome code reviews, but this change is not ready to merge.

This change adds a client factory supports the ability to create clients synchronously.

Fixes #50 

## Problem solved
It is often better to fail to start up an app than to start it with incorrect feature toggle values. Before the changes in this pr there was no feedback from the unleash client if the unleash server was unreachable. This resulted in hacky, race prone workarounds to detect if the unleash client was "ready".

### Change an existing method?
Adding a `.Wait()` or `.Result()` within an existing synchronous block (the https://github.com/Unleash/unleash-client-core/blob/master/src/Unleash/DefaultUnleash.cs constructor, for instance) [could result in deadlocks](https://stackoverflow.com/a/15022170/3430807) during usage. Adding a new factory with an explicitly async signature makes it clear that the call needs to be awaited.
### ExecuteDuringStartup?!
Two workarounds for just this issue were already in the codebase. Unfortunately, they didn't resolve the issue. They have both been removed.
- the `IUnleashScheduledTask.ExecuteDuringStartup` field is intended to signal to `IUnleashScheduledTaskManager` implementations that the task should run at startup. Unfortunately the startup process is synchronous; the default `IUnleashScheduledTaskManager` implementation, `SystemTimerScheduledTaskManager`, simply schedules the task for immediate execution, it does not wait for the task to complete before passing execution back.
- `UnleashServices` contains a test shim that populates the feature toggle collection if it detects an api client has been provided. This works well for https://github.com/Unleash/unleash-client-core/blob/master/tests/Unleash.Tests/ExampleTests.cs where the test requires that the toggles be loaded by the time the unleash client is constructed. With the new client factory, that shim can be removed.

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [X] This change requires a documentation update

# How Has This Been Tested?
//todo

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Firmware version:
* Hardware:
* Toolchain:
* SDK:

# Checklist:

### DIFF
--- a/src/Unleash/ClientFactory/UnleashClientFactory.cs
+++ b/src/Unleash/ClientFactory/UnleashClientFactory.cs
@@ -4,18 +4,26 @@ using System.Threading.Tasks;
 
 namespace Unleash.ClientFactory
 {
+    /// <inheritdoc />
     public class UnleashClientFactory : IUnleashClientFactory
     {
         private UnleashSettings settings { get; }
         private CancellationTokenSource cancellationTokenSource { get; } = new CancellationTokenSource();
 
-
+        ///// <summary>
+        ///// Initializes a new instance of the client factory. 
+        ///// </summary>
+        ///// <param name="settings">Unleash settings</param>
         public UnleashClientFactory(UnleashSettings settings)
         {
             this.settings = settings;
+            this.settings.ScheduleFeatureToggleFetchImmediatly = false;
         }
 
-
+        ///// <summary>
+        ///// Initializes a new instance of Unleash client. 
+        ///// </summary>
+        ///// <param name="SynchronousInitialization">If true, fetch and cache toggles before returning. If false, allow the unleash client schedule an initial poll of features in the background</param>
         public async Task<IUnleash> Generate(bool SynchronousInitialization = false)
         {
             var unleash = new DefaultUnleash(settings);
--- a/src/Unleash/ClientFactory/UnleashClientFactory.cs
+++ b/src/Unleash/ClientFactory/UnleashClientFactory.cs
@@ -1,0 +1,30 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Unleash.ClientFactory
+{
+    public class UnleashClientFactory : IUnleashClientFactory
+    {
+        private UnleashSettings settings { get; }
+        private CancellationTokenSource cancellationTokenSource { get; } = new CancellationTokenSource();
+
+
+        public UnleashClientFactory(UnleashSettings settings)
+        {
+            this.settings = settings;
+        }
+
+
+        public async Task<IUnleash> Generate(bool SynchronousInitialization = false)
+        {
+            var unleash = new DefaultUnleash(settings);
+            if (SynchronousInitialization)
+            {
+                await unleash.services.FetchFeatureTogglesTask.ExecuteAsync(cancellationTokenSource.Token);
+            }
+
+            return unleash;
+        }
+    }
+}
--- a/src/Unleash/ClientFactory/UnleashClientFactory.cs
+++ b/src/Unleash/ClientFactory/UnleashClientFactory.cs
@@ -7,32 +7,20 @@ namespace Unleash.ClientFactory
     /// <inheritdoc />
     public class UnleashClientFactory : IUnleashClientFactory
     {
-        private UnleashSettings settings { get; }
-        private CancellationTokenSource cancellationTokenSource { get; } = new CancellationTokenSource();
-
-        ///// <summary>
-        ///// Initializes a new instance of the client factory. 
-        ///// </summary>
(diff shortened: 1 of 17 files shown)


src/Unleash/ClientFactory/UnleashClientFactory.cs
```diff
@@ -0,0 +1,30 @@
...
+        private CancellationTokenSource cancellationTokenSource { get; } = new CancellationTokenSource();
+
+
+        public UnleashClientFactory(UnleashSettings settings)
+        {
+            this.settings = settings;
+        }
+
+
+        public async Task<IUnleash> Generate(bool SynchronousInitialization = false)
```
AndreasChristianson: Is this parameter needed? Perhaps this should be the default behavior

tests/Unleash.Tests/ClientFactory/SyncStartupUnitTest.cs
```diff
@@ -0,0 +1,29 @@
...
+        public async Task AsyncStartupClientStartup()
+        {
+            var mockApiClient = A.Fake<IUnleashApiClient>();
+            var settings = new MockedUnleashSettings();
+            settings.UnleashApiClient = mockApiClient;
+            var unleashFactory = new UnleashClientFactory(settings);
+
+            IUnleash unleashTask = await unleashFactory.Generate(SynchronousInitialization: true);
+
+            A.CallTo(() => mockApiClient.FetchToggles(string.Empty, A<CancellationToken>.Ignored)).MustHaveHappenedOnceExactly();
```
AndreasChristianson: How do I test that this call was awaited? This only asserts that it was called.

AndreasChristianson:
todo: 
- [ ] update readme
- [ ] document public methods

src/Unleash/Properties/AssemblyInfo.cs
```diff
@@ -1,3 +1,4 @@
 ﻿using System.Runtime.CompilerServices;
 
-[assembly: InternalsVisibleTo("Unleash.Tests")]
\ No newline at end of file
+[assembly: InternalsVisibleTo("Unleash.Tests")]
+[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2")]
```
silviogreuel: What for?
Without it mocking does not work or what?

tests/Unleash.Tests/ClientFactory/SyncStartupUnitTest.cs
silviogreuel: Return some fake toggles and assert the expectations of use, like, unleash should have a flag x.

Besides that, only asserting calls should do the trick.

I think that some tests are missing, like:
- What if it fails or throws exceptions?
- What if `SynchronousInitialization` is set to false? uneleash should have or not toggle x?

src/Unleash/ClientFactory/UnleashClientFactory.cs
silviogreuel: Let maintain the same behavior using new and factory.

src/Unleash/Properties/AssemblyInfo.cs
AndreasChristianson: I was unable to mock `IUnleashApiClient` with fake it easy without letting `DynamicProxyGenAssembly2` see internals.
The fake it easy error message suggested this change

tests/Unleash.Tests/ClientFactory/SyncStartupUnitTest.cs
AndreasChristianson: Added tests showing exceptions bubbling, and `SynchronousInitialization`=false

AndreasChristianson:
I think this is ready for review.
attn @nmiddendorff and @njbatt

AndreasChristianson:
I found a bug that should be addressed before merging this. 
in the api client, [we swallow non-200 responses rather than throwing an exception](https://github.com/Unleash/unleash-client-core/blob/master/src/Unleash/Communication/UnleashApiClient.cs#L45-L55). I want those exceptions to bubble up if `SynchronousInitialization` is true so that apps can cancel startup if unleash is misconfigured.
I'll take a swing at addressing that tonight

PR #64, opened 2020-10-14
Fix timing bug in SystemTimerScheduledTaskManager

### LINKED ISSUE #63: Timing bug in SystemTimerScheduledTaskManager
There is a timing bug in SystemTimerScheduledTaskManager that can prevent the Unleash client from properly scheduling the fetch-feature-toggle-task. This means that the Unleash client will silently not fetch feature toggles.

The problem is the gap in time between when a background task's Timer is created and the Timer is registered in the timers dictionary. 

Specifically: the problem occurs if the callback associated with a task's Timer executes and reaches this line before the Timer is put in the dictionary:
https://github.com/Unleash/unleash-client-dotnet/blob/master/src/Unleash/Scheduling/SystemTimerScheduledTaskManager.cs#L33

Then the callback terminates early, and does not go on to reschedule itself by changing the dueTime for the Timer. Since the period for the Timer is set to infinite, it means that the Timer will never fire again. In other words, the callback executes once and terminates early, without running the task, and without making sure that the Timer will fire again.

einarwh:
# Description

Fixes #63 

The fix changes the way the Timer is used. Instead of changing the Timer's dueTime on each callback, dueTime is only used for the initial delay before firing the callback the first time. It uses the period to specify the interval between each firing of the callback. I think this is how the Timer is supposed to be used.

## Type of change

# How Has This Been Tested?

Verified in our production environment, which runs multiple instances of an ASP.NET Core web application in Azure. 

unleash-bugfix-clip-data-cut

Our configuration looks like this:

```csharp
var unleashSettings = new UnleashSettings
{
  AppName = appName,
  InstanceTag = Environment.MachineName,
  SendMetricsInterval = TimeSpan.FromMinutes(1),
  UnleashApi = new Uri(unleashUrl),
  CustomHttpHeaders = new Dictionary<string, string>
  {
    {"Authorization", clientSecret}
  },
  ScheduledTaskManager = new BugfixTimerScheduledTaskManager(logger)
};

var unleash = new DefaultUnleash(unleashSettings);
```

### DIFF
--- a/src/Unleash/Scheduling/SystemTimerScheduledTaskManager.cs
+++ b/src/Unleash/Scheduling/SystemTimerScheduledTaskManager.cs
@@ -21,58 +21,63 @@ namespace Unleash.Scheduling
         {
             foreach (var task in tasks)
             {
-                var name = task.Name;
+                ConfigureTask(task, cancellationToken);
+            }
+        }
 
-                async void Callback(object state)
+        private void ConfigureTask(IUnleashScheduledTask task, CancellationToken cancellationToken)
+        {
+            var name = task.Name;
+
+            async void Callback(object state)
+            {
+                try
                 {
-                    try
+                    if (!cancellationToken.IsCancellationRequested)
                     {
-                        if (!cancellationToken.IsCancellationRequested)
-                        {
-                            await task.ExecuteAsync(cancellationToken);
-                        }
-                    }
-                    catch (TaskCanceledException taskCanceledException)
-                    {
-                        if (!cancellationToken.IsCancellationRequested)
-                        {
-                            Logger.WarnException($"UNLEASH: Task '{name}' cancelled ...", taskCanceledException);
-                        }
-                    }
-                    catch (Exception ex)
-                    {
-                        Logger.ErrorException($"UNLEASH: Unhandled exception from background task '{name}'.", ex);
-                    }
-                    finally
-                    {
-                        if (cancellationToken.IsCancellationRequested)
-                        {
-                            // Stop the timer.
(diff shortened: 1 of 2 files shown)


einarwh:
There is a regression in that the timer isn't stopped if cancellation has been requested. An option would be to actively stop the timer in that case, e.g. by adding a finally block with something like 
```
if (cancellationToken.IsCancellationRequested)
{
    timer.SafeTimerChange(-1, Timeout.Infinite, ref disposeEnded);
}
```

einarwh:
It seems that even the very conservative bugfix proposed here causes the test to fail: 
https://github.com/Unleash/unleash-client-dotnet/issues/63#issuecomment-707997716

einarwh:
I think a problem in the conservative bugfix scenario is that calling `timer.SafeTimerChange(dueTime, Timeout.InfiniteTimeSpan, ref disposeEnded)` isn't instantaneous. Since the cancellation token is set up to cancel after 75 ms, the task is cancelled before it gets a fair chance to run. Calling `cts.CancelAfter(75)` after `scheduler.Configure` in the test solves the problem I think.

einarwh:
I've updated the bugfix so that it stops the timer on cancellation. I had to adjust the test a bit because the timing details have changed. It's probably a good idea for someone who knows the code better to verify that the test still makes sense.

einarwh:
If you prefer the more conservative fix, I have a branch for that here: 
https://github.com/Unleash/unleash-client-dotnet/compare/master...einarwh:ConservativeTimingBugFix

ivarconr:
I think this solves it quite nice, and also simplifies the code a bit, which is nice.

PR #67, opened 2021-02-09
feat: Exposing the currently loaded FeatureToggle collection outside of the…

### LINKED ISSUE #66: Add Type to feature toggle object and expose the feature toggle collection readonly publicly.
**Is your feature request related to a problem? Please describe.**
Feature parity with other client. Similar to how the node client allows you to get all feature flag definitions: https://github.com/Unleash/unleash-client-node#3-use-unleash 
Also, the endpoint to grab features from the server "client/features" includes the "Type" field which could be useful if we have access to the entire collection.

**Describe the solution you'd like**
Add "Type" as a string to the FeatureFlag class.
Expose a public getter that returns whatever the most recent collection of feature flag definitions exist.  This could be as simple as adding this to DefaultUnleash.cs
`public ICollection<FeatureToggle> FeatureToggles { get { return services.ToggleCollection.Instance.Features; } }`
Doing this would require adjusting the access modifier on some classes like FeatureToggle or would require some sort of mapping to a new public friendly entity.

**Describe alternatives you've considered**
The only other means of doing this would be to hit the api endpoint manually or directly hitting the database. Neither of these are great options.

**Additional context**
Currently, using the net client, a consumer can only grab a flag by name.  Simply exposing the feature flag definition collection would allow a consumer to grab toggles on a type.

brcpar:
… sdk.  Adding "Type" to the FeatureToggle class to be consistent with what comes from the api.  Adjusting tests and access modifiers to account for changes.  This addresses issue#66

# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

I exposed the ICollection<FeatureToggle> of the currently loaded collection of toggles as a public getter so that it can be access outside of the sdk.   I also added Type property to the FeatureToggle class to make consistent with what's coming across the unleash client api.

Fixes #66 

## Type of change

# How Has This Been Tested?

Ran local unleash server and tested access against for new field and also ensure newly exposed property was accessible properly.

# Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,9 @@
 
 ## Introduction
 
-Unleash is a feature toggle system, that gives you a great overview over all feature toggles across all your applications and services. It comes with official client implementations for Java, Node.js and Go.
+Unleash Client SDK for .Net. It is compatible with the
+[Unleash-hosted.com SaaS offering](https://www.unleash-hosted.com/) and
+[Unleash Open-Source](https://github.com/finn-no/unleash).
 
 The main motivation for doing feature toggling is to decouple the process for deploying code to production and releasing new features. This helps reducing risk, and allow us to easily manage which features to enable.
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,6 +7,7 @@ os: Visual Studio 2019
 environment:
   DOTNET_CLI_TELEMETRY_OPTOUT: true
   DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
+  CAKE_SETTINGS_SKIPPACKAGEVERSIONCHECK: true
 
 install:
   - choco install gitversion.portable --version 4.0.0 -y
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,7 +20,7 @@ before_build:
   - ps: gitversion /l console /output buildserver 
 
 build_script:
- - ps: .\build.ps1 -Target "AppVeyor"
+ - ps: .\build.ps1 --Target "AppVeyor"
  
 test: off
 
--- a/build.ps1
+++ b/build.ps1
@@ -220,13 +220,13 @@ if (!(Test-Path $CAKE_EXE)) {
 
 # Build Cake arguments
 $cakeArguments = @("$Script");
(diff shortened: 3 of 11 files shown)


ivarconr:
Thanks @brcpar. 

Do you happen to understand why AppVeyor fails with:

```
Error: Unknown command 'AppVeyor'.
       build.cake -target=AppVeyor
                          ^^^^^^^^ No such command
```

brcpar:
I'm not sure what could have caused that.  That appears to be something with the CI build.  I built using the dotnet cli fine before PRing.  I can try to take a look at what specifically could be failing the cake build.

brcpar:
I haven't had time to debug it yet, but I pulled master and tried to run the cake build and it too fails.  It appears to be something already in master causing this. I might have more time to poke around at it in a few hours.

brcpar:
Actually, after just running the ps command, I noticed if you specify target to being default, it doesn't the same thing and the command line it generates which fails is:
`.\tools\Cake\Cake.exe build.cake -target=default`
However, this appears to build:
`.\tools\Cake\Cake.exe build.cake --target=default`

I'm not very savvy with cake builds, but it appears as though build.ps1 needs to be updated to pass `--target` rather than `-target` in the script generation at the bottom.

@ivarconr, that's just my guess.

ivarconr:
Ok, fixed that on master, do you mind rebase your branch against master?

brcpar:
I rebased your changes into my branch and it's now failing with an error on gitversion.  I'm not familiar enough with gitversion to say why it's failing.  Locally, gitversion produces output without any exceptions.

ivarconr:
The gitversion is called as a result of an argument error which was a result of changing the arguments (from -target it --target). I will try to have a deep look at this tomorrow. Funny the PR i created seems to work just fine.

ivarconr:
Ok, sorry for the back an forth (I had lost access to the appveyor project), but restored that and now fixed the build. Would you mind rebase another time from master? 

Except from that this PR makes sense, and is good to merge!

ivarconr (approved):

PR #84, opened 2022-01-05
feat: Add support for bootstrapping Unleash

### LINKED ISSUE #83: feat: Add support for bootstrapping feature toggles
Improve resilience by implementing the Bootstrap feature of Unleash. Feature writeup:
https://www.getunleash.io/blog/bootstrap-unleash-java-sdk-to-improve-the-resilience

Reference implementation in Java SDK:
https://github.com/Unleash/unleash-client-java/pull/131

Primary goals:
- Add a property to UnleashSettings for configuring an `IToggleBootstrapProvider `to use
- Add an implementation of `IToggleBootstrapProvider` that can read from a specified file path
- Add an implementation of `IToggleBootstrapProvider` that can GET from a specified url
- Read from `IToggleBootstrapProvider` only when there is no backup file or the backup files ToggleCollection is empty

daveleek:
# Description

- Adds unittests for existing implementation of CachedFileLoader
- Fixes recent build errors by pinning Cake version
- Adds support for bootstrapping unleash from file and url

This pull requests goal is to implement the feature described in #83; bootstrapping feature toggles.
It provides two bootstrap toggle providers, one that loads from a filepath, and one that GETs from a url. A configuration option for `ToggleBootstrapProvider` is added to the UnleashSettings, and the CachedFileLoader is modified to add support for invoking bootstrapping at startup if loading from backup fails, or if told to override backup if bootstrapped toggles are available. (Override defaults to `true`, set `BootstrapOverride` on UnleashSettings to `false` to disable override).

Fixes #83 

## Type of change

# How Has This Been Tested?

For now this has been tested by adding unittests to verify existing code before modifying, and adding unittests to verify new feature additions and modifications

# Checklist:


src/Unleash/Internal/CachedFilesLoader.cs
```diff
@@ -49,6 +51,7 @@ public CachedFilesResult EnsureExistsAndLoad()
             if (result.InitialToggleCollection == null)
             {
                 result.InitialETag = string.Empty;
+                result.InitialToggleCollection = toggleBootstrapHandler.Read();
```
ivarconr, line 54: I know this aligns with the Java SDK. But we actually think it may be a mistake to not allow it to override the locally backup file. The default should be that bootstrap can override a local cache, but it should be an option the user can set. 

`boostrapOverride = true` should default to true, and means that bootstrap will override InitialToggleCollection, given that the bootstrap collection is valid and not an empty collection.

src/Unleash/Internal/ToggleBootstrapHandler.cs
```diff
@@ -0,0 +1,37 @@
...
+        }
+
+        public ToggleCollection Read()
+        {
+            if (toggleBootstrapProvider == null)
+                return null;
+
+            var bootstrapContent = toggleBootstrapProvider.Read();
+            var bootstrapContentBuffer = encoding.GetBytes(bootstrapContent ?? "");
+            using (var bootstrapStream = new MemoryStream(bootstrapContentBuffer))
```
sighphyre, line 31: The bootstrap provider seems to return a string, which means that it's already read a byte stream from a socket/filehandle and converted that to a string, do we have convert it back to byte stream?

src/Unleash/Utilities/ToggleBootstrapFileProvider.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        internal ToggleBootstrapFileProvider(string filePath, IFileSystem fileSystem)
+        {
+            this.filePath = filePath;
+            this.fileSystem = fileSystem;
+        }
+
+        public string Read()
+        {
+            if (!fileSystem.FileExists(filePath))
```
sighphyre, line 27: This might be a personal thing (I may have spent too much time in Python) but I would probably remove this guard here for two reasons:

1) Someone has to handle the potential IOException anyway since the last line of this function reads a file and that can potentially fail for reasons outside of the program's control
2) The incredibly rare race condition where a file is wiped between checking for its existence and obtaining a filehandle is taken care of by the previous case anyway

src/Unleash/Utilities/FetchingToggleBootstrapUrlFailedException.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Net;
+using System.Text;
+
+namespace Unleash.Utilities
+{
+    public class FetchingToggleBootstrapUrlFailedException : Exception
```
sighphyre, line 8: Is it possible/sensible to make this inherit from UnleashException? It seems like everything else in the code uses that and it'd be nice for consumers to only have to worry about a single exception type (if that's rational, you know your code best)

src/Unleash/Utilities/ToggleBootstrapUrlProvider.cs
```diff
@@ -0,0 +1,56 @@
...
+        public ToggleBootstrapUrlProvider(string path, HttpClient client, bool throwOnFail = false)
+        {
+            this.path = path;
+            this.client = client;
+            this.throwOnFail = throwOnFail;
+        }
+
+        public string Read()
+        {
+            return Task.Run(() => FetchFile()).Result;
```
sighphyre, line 34: Seems like a terrible shame to go through the async/await dance only to synchronize and block. Is this so we can use this in a constructor?

src/Unleash/Utilities/ToggleBootstrapUrlProvider.cs
```diff
@@ -0,0 +1,56 @@
...
+            using (var request = new HttpRequestMessage(HttpMethod.Get, path))
+            {
+                using (var response = await client.SendAsync(request, cancellationTokenSource.Token).ConfigureAwait(false))
+                {
+                    if (!response.IsSuccessStatusCode)
+                    {
+                        if (throwOnFail)
+                            throw new FetchingToggleBootstrapUrlFailedException("Failed to fetch feature toggles", response.StatusCode);
+
+                        return string.Empty;
```
sighphyre, line 48: I don't know the code here well enough to say that this must happen but I've spent too much of my life hunting down exceptions that got swallowed. Is it rational/possible to log this error case here if it doesn't bubble?

README.md
```diff
@@ -240,6 +240,44 @@ By default unleash-client fetches the feature toggles from unleash-server every
...
+#### ToggleBootstrapFileProvider
+* Unleash comes with a `ToggleBootstrapFileProvider` which implements the `IToggleBootstrapProvider` interface.
+* Constructor takes the `path` parameter used to find the JSON source file
+
+#### ToggleBootstrapUrlProvider
+* Unleash also comes with a `ToggleBootstrapUrlProvider` which implements the `IToggleBootstrapProvider` interface.
+* Fetches JSON from a webaddress using `HttpMethod.Get`
+* The constructor takes the `path` parameter as the webaddress for the JSON
+* The constructor takes an optional `client` `HttpClient` parameter, use when reusing clients or configuring custom headers
+* The constructpr also takes an optional `throwOnFail` `bool` parameter that defaults to `false`. Set it to `true` if you need HTTP 404:s and 500:s etc to throw
```
sighphyre, line 279:
```suggestion
* The constructor also takes an optional `throwOnFail` `bool` parameter that defaults to `false`. Set it to `true` if you need HTTP 404:s and 500:s etc to throw
```

src/Unleash/Utilities/ToggleBootstrapUrlProvider.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+        private readonly CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
+        private readonly HttpClient client;
+        private readonly string path;
+        private readonly bool throwOnFail;
+
+        public ToggleBootstrapUrlProvider(string path, bool throwOnFail = false)
+        {
+            this.path = path;
+            client = new HttpClient();
```
sighphyre, line 21: The HttpClient in dotnet can (and probably should) be reused, it's thread safe but it isn't safe to instantiate over and over again. We've had at least one issue in the past with socket exhaustion due to this (https://github.com/Unleash/unleash-client-dotnet/pull/41/files). The way the client is used here is probably safe. It's almost certainly safe. But I would sleep better at night if this were reused/cached/static

sighphyre (changes requested):
Looking very cool! I have some comments and thoughts - please feel free to tell me I'm wrong in places and mark things with a resolution and response!

src/Unleash/Internal/CachedFilesLoader.cs
daveleek: Ok good stuff, that makes sense! i'll add some more tests and implement this!

src/Unleash/Internal/ToggleBootstrapHandler.cs
daveleek: Yeah this is actually one of the things i've not been 100% happy with implementation wise, and want to revisit before completion, so i'm happy to get comments on this!

The interface used to deserialize JSON only takes a Stream input, so in order not to grow the scope too much, and be close to the java reference sdk, provide an easy interface for others to implement + skip some hoops in testing, this is the tradeoff i opted for to keep moving.

The two options as i see them are to convert back to exposing the original streams and expect the caller to dispose properly, or build an overload in the JSON deserialization abstraction and reflection code to handle a string parameter. I might be leaning ever so slightly in preference towards the latter for the reasons mentioned, but will happily take your thoughts on that!

Sorry for the long reply

src/Unleash/Utilities/ToggleBootstrapFileProvider.cs
daveleek: Yeah i think you're right. If you code explicitly for a bootstrapfile to be used you'll probably want an operational exception to let you know when it doesn't exist.

src/Unleash/Utilities/FetchingToggleBootstrapUrlFailedException.cs
daveleek: Yup defo, good catch!

src/Unleash/Utilities/ToggleBootstrapUrlProvider.cs
daveleek: Yeah, this is done in the UnleashServices constructor, and while we could probably move this all the way up to the constructor and keep it async down on from there i didn't want to casually introduce that big a change. The other thing i took into consideration is that this most likely occurs only once during startup, and freeing up threads while waiting on IO might not be a priority at that point

src/Unleash/Utilities/ToggleBootstrapUrlProvider.cs
daveleek: Yeah, there will be logging. It's on the backlog before considering myself done! Will make sure to include logging here as well! Do you have an opinion on throwOnFail or not? When you mentioned wanting things to fail in the fileprovider it got me thinking about this case as well

README.md
daveleek: Thanks 😬

src/Unleash/Utilities/ToggleBootstrapUrlProvider.cs
daveleek: Yeah, there's an overload for providing an HttpClient already, specificly to allow reuse (but also to allow configuring custom headers and makes testing easier), but i could definately remove this particular constructor and require a client as parameter. I could also looking at how easy it would be to just reuse the client used in the communication with the Unleash API

daveleek:
Thank you @sighphyre i have left some replies and return questions

src/Unleash/Utilities/ToggleBootstrapUrlProvider.cs
sighphyre: Yeah I hear you, that sounds like a rather big change. I'm not very familiar with this SDK and it seems to want to do things in different ways from the others. I think we should just accept this and potentially look at a refactor after (if we even want to, I think your arguments for leaving it alone are good)

src/Unleash/Utilities/ToggleBootstrapUrlProvider.cs
sighphyre: Thank you! You're a star

I'm not sure I have a useful opinion on the throwOrFail so take this for what it's worth haha: I''m not sure I see a need for it to be honest; I don't see a usecase for allowing an exception to bubble (and potentially take down someone else's code) with an unhandled exception. But at the same time it defaults to false and if it's logged I'm happy. The only relevant comment I have is that I think it should inherit from the UnleashException base class so that consumers can catch that type and be safe. Personally I would have named this "failQuietly" but that feels like an unhelpful super nitpicky comment so I just kept quiet

src/Unleash/Utilities/ToggleBootstrapUrlProvider.cs
sighphyre: If you wouldn't mind looking that would be awesome. If it's a huge amount of work then it's probably not worth it and I'll come back to this when it's done (and the other SDKs aren't keeping me busy!) and swap it out later. It's not urgent, like I said, this requires a consumer to do silly things before it goes wrong

src/Unleash/Internal/ToggleBootstrapHandler.cs
sighphyre: Ha! As I was looking at this I was thinking to myself "I bet this is because this interface is unwieldy"

So I don't have a clear, simple or obvious solution to this issue, but I do have some thoughts and I'm going to preface this by saying this is the Rust dev in me saying this, not the C# dev so if it makes you uncomfortable, just ignore me. Right now I feel like the ToggleBootstrapHandler doesn't do very much, it seems to just be responsible to returning a result of a IToggleBootstrapProvider and parsing that string to a ToggleCollection. That abstraction seems to be leaking a little for two reasons: 1) The ToggleBootstrapHandler knows about FileSystems (does it need to?) and 2) We have a dance where we're doing string/bytestream conversions. I don't think we need this amount of abstraction, I would just collapse the responsibility for calling a JSON parser into the BootStrapProviders (right now they don't actually have anything to do with Toggles, they read a file and a socket and they could do that job against any datasource really). This also means there's no real validation that what the BootstrapProvider returns will actually parse to a ToglleCollection, so this allows us to be suuuuper careful about error handling (yup, the Rust dev talking again). Right now there's a bunch of exceptions that can bubble from parsing a JSON bytestream and we're not doing anything with them

(I see your long reply and I raise you this one!)

src/Unleash/Utilities/ToggleBootstrapUrlProvider.cs
amj1985: Task.Run(() => FooBar()) Should be used in a CPU bound operation not in a I/O Bound operation, also, is better to use GetAwaiter().GetResult() than .Result directly because in case of failure, the first one will propagate the exceptions and the second one will wrap in a AggregateException.

daveleek: @amj1985 Thank you, yes indeed. The reason i opted for Task.Run to avoid deadlocks. I'll admit to not being the ultimate brains on this topic, but my understanding is that GetAwaiter().GetResult() can cause deadlocks in things like ASP.NET (not Core) and other single threaded use cases that can't neccessarily be ruled out

src/Unleash/Utilities/ToggleBootstrapUrlProvider.cs
daveleek: Cool, thank you!

src/Unleash/Utilities/ToggleBootstrapUrlProvider.cs
daveleek: Yup, i'll have a look and see! Thanks!

src/Unleash/Utilities/ToggleBootstrapUrlProvider.cs
amj1985: Both cases .Result .GetAwaiter().GetResult() cause the potential for deadlock and thread pool starvation issues. It should be avoided and be implemented using async all the way, but it involves a big refactor, my recommendation is to use GetAwaiter().GetResult(), if the process throws
a FileNotFoundException will get the actual exception instead of an AggregateException

daveleek: @amj1985 ah i see what you mean now, apologies, i missread. Task.Run(...).GetAwaiter().GetResult() vs .Result. Yeah reasonable request, i'll fix that!

sighphyre: @amj1985 I think we should lean towards discussing a refactor soon. The fact that this is currently hard to implement without dissecting technical internals of dotnet runtimes suggests that its time! (I'm keen and willing to help here)

src/Unleash/Utilities/ToggleBootstrapUrlProvider.cs
```diff
@@ -0,0 +1,69 @@
...
+                        Logger.Trace($"UNLEASH: Error {response.StatusCode} from server in 'ToggleBootstrapUrlProvider.{nameof(FetchFile)}': " + error);
+
+                        if (throwOnFail)
+                            throw new FetchingToggleBootstrapUrlFailedException("Failed to fetch feature toggles", response.StatusCode);
+
+                        return null;
+                    }
+
+                    var togglesResponseStream = await response.Content.ReadAsStreamAsync();
+                    return jsonSerializer.Deserialize<ToggleCollection>(togglesResponseStream);
```
sighphyre, line 64: It might be worth wrapping these two lines in some kind of error handling. Since we have a neat little option to configure whether or not the SDK will raise an exception when bootstrapping fails, I'd be quite surprised to set this to false and then get an exception. ReadAsStreamAsync can fail in a few cases, the notable one for me is that when Unleash servers are extremely overloaded, they may choose to close the connection early, this will raise an exception here and crash(?) the parent process. The deserialize can also fail, that's probably more on the user for pointing the bootstrap to a broken data source so open to opinions on that one but I do think we should trap the first case

sighphyre (commented):
This looks really cool. If we can wrap up the last comment I'd be super happy to merge this!

src/Unleash/Utilities/ToggleBootstrapUrlProvider.cs
daveleek: @sighphyre Ok fixed. I figured wrap both, if we're only trapping the first one then we should name the parameter accordingly to avoid confusion, so opted to trap both cases. Personal opinion, but open to discussion.

sighphyre (approved):

src/Unleash/Utilities/ToggleBootstrapUrlProvider.cs
sighphyre: I like it!

PR #104, opened 2022-07-06
Feature/global segments 103

daveleek:
# Description

This PR attempts to solve two things: passing the supported spec version header along with the fetch toggles api request, and support for global segments.

Fixes # (issue)
#103

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

For now only tested  by running the full spec test suite, unit tests are being added
- [x] Added and ran specific unit tests
- [x] Ran feature spec tests, and crafted a heavier unit test in a test branch for stressing it with data

# Checklist:

### DIFF
--- a/src/Unleash/DefaultUnleash.cs
+++ b/src/Unleash/DefaultUnleash.cs
@@ -214,6 +214,10 @@ namespace Unleash
                         yield return constraint;
                     }
                 }
+                else
+                {
+                    yield return null;
+                }
             }
         }
 
--- a/src/Unleash/DefaultUnleash.cs
+++ b/src/Unleash/DefaultUnleash.cs
@@ -107,7 +107,7 @@ namespace Unleash
             else
             {
                 var enhancedContext = context.ApplyStaticFields(settings);
-                enabled = featureToggle.Strategies.Any(s => GetStrategyOrUnknown(s.Name).IsEnabled(s.Parameters, enhancedContext, s.Constraints));
+                enabled = EnabledByAnyStrategy(featureToggle, enhancedContext);
             }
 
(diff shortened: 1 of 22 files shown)


src/Unleash/DefaultUnleash.cs
```diff
@@ -202,6 +202,34 @@ private IStrategy GetStrategyOrUnknown(string strategy)
...
+            {
+                var segments = new List<Segment>(activationStrategy.Segments.Count);
+                foreach (var segmentId in activationStrategy.Segments)
+                {
+                    var segment = services.ToggleCollection.Instance.GetSegmentById(segmentId);
+                    if (segment == null)
+                    {
+                        return false;
+                    }
+                    segments.Add(segment);
```
sighphyre, line 217: I think this is probably fine but the segments list might get largish. IIRC a list in C# has a memory size of N * sizeof T where T is the generic the list holds. This might be a lot of memory at load in a fairly tight loop. Might be worth checking what this does the process memory when there are 10K segments

daveleek: Yeah that's a very good point, and exactly the type of feedback i was soliciting. I'm doing an IEnumerable rewrite of this, i'll see if i can run a few simple benchmarks of it

sighphyre: Amazing, that sounds great! There are a few scripts in Unleash server in the /perf folder to load a bunch of data into the server for testing exactly this, they were pretty useful for testing this when building the feature on the other SDKs

sighphyre (approved):
LGTM :)

PR #110, opened 2022-10-18
fix: wrap IO tasks in try catch so we can fail more gracefully

nunogois:
# Description

Wraps IO writes in try catch so we can fail more gracefully: https://unleash-community.slack.com/archives/CGP2MCHPF/p1666073059611789

## Type of change

# Checklist:

### DIFF
--- a/src/Unleash/Internal/CachedFilesLoader.cs
+++ b/src/Unleash/Internal/CachedFilesLoader.cs
@@ -1,9 +1,13 @@
+using System.IO;
+using Unleash.Logging;
+using Unleash.Scheduling;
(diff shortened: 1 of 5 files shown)


sighphyre (changes requested):
Looks good!

Could we include a test for this before we merge it? Mostly to validate that it was an issue that's now fixed

And I'm not going to block the PR for this but your autoformatter seems to have decided that two spaces is the way to go here, I'd rather not change the style of this file if we can avoid it. If you're up for restoring the formatting that'd be nice but not a blocker

sighphyre (approved):
Yay nice!

Issue #108, opened 2022-09-15
feat: Event system


daveleek:
**Is your feature request related to a problem? Please describe.**
The task to implement [Impression Event Data](https://docs.getunleash.io/advanced/impression-data#impression-event-data)

**Describe the solution you'd like**
To add support for the Impression Event Data feature to the Unleash .NET SDK

**Describe alternatives you've considered**

**Additional context**
I'll start the discussion below with a couple of suggestions for the client facing part of implementation as requested
@sighphyre @chriswk @ivarconr

daveleek:
# Event system client facing API

Proposals for the way client code interacts with the event system. 
There's quite a few ways to go about this in .NET, I'll propose the ones I've seen used the most. 
Obviously feel free to pick on namings, mix and match implementations, suggest other approaches etc.

Other things of interest to discuss might be async invocation vs synchronous invocation of message handlers/callbacks, should we catch+log+rethrow? catch+log+swallow? etc.

## 1. Configured in a separate class on property in UnleashSettings initializer

```csharp
var settings = new UnleashSettings()
{
    AppName = "event-system-client-facing-examples",
    InstanceTag = "instance z",
    UnleashApi = new Uri("http://localhost:4242/api/"),
    CustomHttpHeaders = new Dictionary()
    {
      {"Authorization","API token" }
    },

    //==== This is the part we're adding ====/
    Events = new EventsConfig
    {
        // A. Callback option
        Impression = evt => {  Console.WriteLine($"Impression: { evt.EventType }"); },
        // B. Class/Interface option
        Impression = new ImpressionEventHandler(),
    }
};
```
### Infrastructure stuff

```csharp
public class EventsConfig
{
    // A. Callback option
    public Action<ImpressionEvent> Impression { get; set; }

    // B. Class/Interface option
    public IEventHandler<ImpressionEvent> Impression { get; set; }
}

public class UnleashSettings
{
    //....

    // Setting added as property and set on initialization
    public EventsConfig Events { get; set; }
}

public class ExpressionEventHandler : IEventHandler<ImpressionEvent>
{
    public void Handle(ImpressionEvent evt)
    {
        Console.WriteLine($"Impression: { evt.EventType }");
    }
}

```

## 2. Register-method on UnleashSettings

Generics, marker interface, and inheritance gives us explicit method invoking for client and provides us with the metadata we need to keep track of handlers..

Should this method be called On()? Register()? Any other name you'd rather want?

```csharp
[...]

PR #113, opened 2022-10-25
Feature: event system and impression data events

daveleek:
# Description
Aims to add an event system to the SDK and a first implementation for impression data/events: https://docs.getunleash.io/advanced/impression-data#impression-event-data

Fixes # (issue)
#108 

## Type of change

# How Has This Been Tested?
- [x] ran all the existing unit tests
- [x] ran the cake script testing
- [x] added more unit tests and ran them
- [x] explicitly tested serialization/deserialization

# Checklist:

### DIFF
--- a/src/Unleash/DefaultUnleash.cs
+++ b/src/Unleash/DefaultUnleash.cs
@@ -65,7 +65,7 @@ namespace Unleash
         /// <inheritdoc />
         public ICollection<FeatureToggle> FeatureToggles => services.ToggleCollection.Instance.Features;
 
-        private EventCallbackConfig EventConfig => new EventCallbackConfig();
+        private EventCallbackConfig EventConfig { get; set; }
 
         /// <inheritdoc />
         public bool IsEnabled(string toggleName)
@@ -229,15 +229,18 @@ namespace Unleash
             if (callback == null)
             {
                 Logger.Error($"UNLEASH: Unleash->ConfigureEvents parameter callback is null");
+                return;
             }
 
             try
             {
-                callback(EventConfig);
+                var evtConfig = new EventCallbackConfig();
+                callback(evtConfig);
+                EventConfig = evtConfig;
             }
             catch (Exception ex)
             {
-                Logger.Error($"UNLEASH: Unleash->ConfigureEvents executing callback threw exception {ex.Message}");
+                Logger.Error($"UNLEASH: Unleash->ConfigureEvents executing callback threw exception: {ex.Message}");
             }
         }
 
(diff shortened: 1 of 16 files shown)


src/Unleash/DefaultUnleash.cs
```diff
@@ -235,6 +243,52 @@ private IEnumerable<Constraint> ResolveConstraints(ActivationStrategy activation
...
+                return;
+            }
+
+            try
+            {
+                var evtConfig = new EventCallbackConfig();
+                callback(evtConfig);
+                EventConfig = evtConfig;
+            }
+            catch (Exception ex)
```
sighphyre, line 260: Is there something that can be raised here? This looks pretty harmless to me

src/Unleash/DefaultUnleash.cs
```diff
@@ -235,6 +243,52 @@ private IEnumerable<Constraint> ResolveConstraints(ActivationStrategy activation
...
+                {
+                    Type = type,
+                    Context = context,
+                    EventId = Guid.NewGuid().ToString(),
+                    Enabled = enabled,
+                    FeatureName = name,
+                    Variant = variant
+                });
+            }
+            catch (Exception ex)
```
sighphyre, line 286: Same as above, I can't see an obvious exception that could be caught here, is this just being defensive?

sighphyre (commented):
This is pretty cool, just a small question about error handling but looks pretty darn good to me

src/Unleash/DefaultUnleash.cs
daveleek: This particular one is probably very harmless, it's the callback where you configure all the handlers for the various events. But anything started in here that fails on a file lock or whatever will crash the startup of Unleash

src/Unleash/DefaultUnleash.cs
daveleek: Here I'm trapping event emitting, if your logger fails for some reason for instance, that won't crash harder than an error in the log

daveleek:
Thanks @sighphyre for review and questions. I've attempted a few more details in response. Perhaps this is a code area that could do with some comments 🤔 

I'm trapping exceptions here since these two callbacks are code coming from outside Unleash, but I'm open for anything :)

src/Unleash/DefaultUnleash.cs
sighphyre: I can stand by that!

src/Unleash/DefaultUnleash.cs
sighphyre: Sure, I can live with that, seems polite to log an error if something super unexpected happens here

sighphyre (approved):
Looks good to me

PR #118, opened 2022-12-04
feat: add support for ip cidr in remote address strategy

daveleek:
# Description

Adds support for IP CIDR (IPv4 & IPv6) in RemoteAddressStrategy

Fixes # (issue)

## Type of change

# How Has This Been Tested?

Added tests specificly for the added feature. Ran all the existing tests

# Checklist:

### DIFF
--- a/src/Unleash/Utilities/IPCIDRAddressRange.cs
+++ b/src/Unleash/Utilities/IPCIDRAddressRange.cs
@@ -17,6 +17,9 @@ namespace Unleash.Utilities
             cidrCount = int.Parse(ipAndCidrPair[1]);
             ipAddress = IPAddress.Parse(ipAndCidrPair[0]);
             baseIPBytes = ipAddress.GetAddressBytes();
+
+            if (cidrCount > (baseIPBytes.Length * 8))
+                cidrCount = baseIPBytes.Length * 8;
         }
 
         public bool Contains(IPAddress remoteAddress)
--- a/src/Unleash/Utilities/IPCIDRAddressRange.cs
+++ b/src/Unleash/Utilities/IPCIDRAddressRange.cs
@@ -8,15 +8,14 @@ namespace Unleash.Utilities
     internal class IPCIDRAddressRange
     {
         private int cidrCount;
-        private IPAddress ipAddress;
-        private byte[] baseIPBytes;
+        private IPAddress baseIPAddress;
 
         public IPCIDRAddressRange(string address)
         {
             var ipAndCidrPair = address.Split('/');
             cidrCount = int.Parse(ipAndCidrPair[1]);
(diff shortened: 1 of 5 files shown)


src/Unleash/Utilities/IPCIDRAddressRange.cs
```diff
@@ -0,0 +1,61 @@
...
+            // We've reached the end of the CIDR subnet mask
+            if (remaining == 0)
+                return true;
+
+            // Blank out all variable bits so we can compare the bytes to each other
+            for (var shift = 0; shift + remaining < 8; shift++)
+            {
+                byte mask = (byte)(1 << shift);
+                remoteBytes[currentByte] &= (byte)~mask;
+                baseIPBytes[currentByte] &= (byte)~mask;
```
sighphyre, line 54: Doesn't this mutate the object's field in place? It seems safe in this code since we're discarding these objects after a single use but at some point someone's going to try reuse this code and get burned (I would cry if I invoked a method called `contains` and that mutated the parent object). I want to say it's safe, cheap and easy to take a clone of `baseIPBytes` at the top of the method and use that rather. WDYT?

tests/Unleash.Tests/Strategy/RemoteAddressStrategyTest.cs
```diff
@@ -62,5 +62,65 @@ public void test(string actualIp, string parameterstring, bool expected)
...
+        [Test]
+        public void GuardsAgainstTooHighCIDR()
+        {
+            var range = "73.125.227.1/59";
+            var input = "73.125.227.1";
+            var context = UnleashContext.New().RemoteAddress(input).Build();
+            var parameters = setupParameterMap(range);
+            strategy.IsEnabled(parameters, context).Should().BeTrue();
+        }
     }
```
sighphyre, line 125: The tests look good to me! I suggest adding another one check that the same strategy can be invoked multiple times and returns the same result, just to be defensive about the mutations we need in `IPCIDRAddressRange` so we don't get burned later

sighphyre (commented):
I think this looks good, left a small suggestion around some mutation stuff, open for comments on my comments

src/Unleash/Utilities/IPCIDRAddressRange.cs
daveleek: Yeah i think i'll just keep the ipaddress instead of the bytes result, and then just ask for the bytes when i need them 👍

tests/Unleash.Tests/Strategy/RemoteAddressStrategyTest.cs
daveleek: Good thinking! adding!

chriswk (approved):
LGTM. Nothing like a little /junk CIDR mask :)

PR #120, opened 2023-01-09
docs: merge content from docs.getunleash.io

thomasheartman:
This change merges content from docs.getunleash.io into the readme

Most of the content already existed in the readme, so the changes are
primarily rewordings and clarifications.

There's also a few cleanup changes:

- Promote a fair few headings. It seems like a lot of the headings were
mistakenly nested under "Getting started", so I have promoted them so
that they're now they're own proper subsections.
- Updated links to the docs.getunleash.io articles

## Background

We're making a change to the documentation on docs.getunleash.io. Keeping the docs in sync with the SDKs and their readmes is difficult and a lot of manual work, so we've decided to move to generating the SDK documentation from the project readmes. Moving forward, all SDK-specific documentation will be replaced with the SDKs' respective github readmes. (Refer to https://github.com/Unleash/unleash/pull/2809 for more background.)

This PR aims to update this project readme to make sure that no information is lost in this content.

## Type of change

- [x] This change is a documentation update

# How Has This Been Tested?
N/A

# Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ If you create more than 10 instances, Unleash will attempt to log warnings about
 To create a new instance of Unleash you need to create and pass in an `UnleashSettings` object.
 
 When creating an instance of the Unleash client, you can choose to do it either **synchronously** or **asynchronously**.
-The SDK will synchronize with the Unleash API on initialization, so it can take a moment for the client to reach the correct state.
+The SDK will synchronize with the Unleash API on initialization, so it can take a moment for the it to reach the correct state. With an asynchronous startup, this would happen in the background while the rest of your code keeps executing. In most cases, this isn't an issue. But if you want to **wait until the SDK is fully synchronized**, then you should use the configuration explained in the [synchronous startup](#synchronous-startup) section.
 This is usually not an issue and Unleash will do this in the background as soon as you initialize it.
 However, if it's important that you do not continue execution until the SDK has synchronized, then you should use the configuration explained in the [synchronous startup](#synchronous-startup) section.
 
--- a/README.md
+++ b/README.md
@@ -35,23 +35,41 @@ Extendable architecture
 - Inject your own implementations of key components (Json serializer, background task scheduler, http client factory)
 
 ## Getting started
-Install the latest version of `Unleash.Client` from [nuget.org](https://www.nuget.org/packages/Unleash.Client/).
+
+### Install the package
+
+Install the latest version of `Unleash.Client` from [nuget.org](https://www.nuget.org/packages/Unleash.Client/) or use the `dotnet` cli:
+
+``` bash
+dotnet add package unleash.client
+```
 
 ### Create a new a Unleash instance
 
-It is easy to get a new instance of Unleash. In your app you typically *just want one instance of Unleash*, and inject that where you need it. You will typically use a dependency injection frameworks to manage this. 
+---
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -35,23 +35,41 @@ Extendable architecture
...
+**⚠️ Important:** In almost every case, you only want a **single, shared instance of the Unleash  client (a *singleton*)** in your application . You would typically use a dependency injection framework to inject it where you need it. Having multiple instances of the client in your application could lead to inconsistencies and performance degradation.
+
+If you create more than 10 instances, Unleash will attempt to log warnings about your usage.
+
+---
+
+To create a new instance of Unleash you need to create and pass in an `UnleashSettings` object.
+
+When creating an instance of the Unleash client, you can choose to do it either **synchronously** or **asynchronously**.
+The SDK will synchronize with the Unleash API on initialization, so it can take a moment for the client to reach the correct state.
```
sighphyre, line 60: These two lines aren't all that clear to me that they're talking about what happens when you're doing async init. Maybe something like: 

"When initialised asynchronously, the SDK will synchronize with the Unleash API on initialization, so it can take a moment for the client to reach the correct state.
This is usually not an issue and Unleash will do this in the background as soon as you initialize it."

README.md
```diff
@@ -145,10 +203,11 @@ The .Net client comes with implementations for the built-in activation strategie
...
 - FlexibleRolloutStrategy
 
-Read more about the strategies in [activation-strategy.md](https://docs.getunleash.io/user_guide/activation_strategy).
+Read more about the strategies in [the activation strategy reference docs](https://docs.getunleash.io/reference/activation-strategies).
 
-#### Custom strategies
-You may also specify and implement your own strategies. The specification must be registered in the Unleash UI and you must register the strategy implementation when you wire up unleash. 
+### Custom strategies
+
+You may also specify and implement your own strategies. The specification must be registered in the Unleash UI and you must register the strategy implementation when you wire up unleash.
```
sighphyre, line 210: Does this deserve a link to https://docs.getunleash.io/reference/custom-activation-strategies?

README.md
thomasheartman: Hmm, it's not necessarily about async init, though. Just about what happens when you start it. Whether you wait for it or not is up to you 🤔  How about this:

```suggestion
The SDK will synchronize with the Unleash API on initialization, so it can take a moment for the it reach the correct state. With an asynchronous startup, this would happen in the background while the rest of your code keeps executing. In most cases, this isn't an issue. But if you want to **wait until the SDK is fully synchronized**, then you should use the configuration explained in the [synchronous startup](#synchronous-startup) section.
```

I'm a bit unsure about saying the "sdk will synchronize with the API", because I'm afraid it can be confusing when talking about synchronous and asynchronous startups (a lot of syncing here), so I'm open to rewording suggestions.

README.md
thomasheartman: Sure does! Addressed in 75bf7fe

README.md
sighphyre: Maybe I'm too close to this and I have too much context already but I think this is as clear as it gets. I think the rewording here is pretty solid so I'm happy to approve.

Small detail though: "so it can take a moment for the it reach the correct state" did you mean to remove the "the" here?

thomasheartman: D'ya mean the "to"? And no, I did not. added back in now 😅

sighphyre (approved):
LGTM

PR #124, opened 2023-02-21
docs: Remove references to `InstanceTag`

thomasheartman:
## What

This change removes references to `InstanceTag` in the examples.

## Why

1. This property doesn't need to be set. The SDK will generate a default if you don't provide it. Removing it simplifies the examples.

2. This property is never explained anywhere in the readme, so it can
be hard to know what it does and what it's used for. Additionally, the
name `InstanceTag` is different from the `InstanceId` that other SDKs
use, making it confusing even to people somewhat familiar with Unleash.

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,7 +267,6 @@ protected void Application_BeginRequest(object sender, EventArgs e)
 var settings = new UnleashSettings()
 {
     AppName = "dotnet-test",
-    InstanceTag = "instance z",
     UnleashApi = new Uri("http://unleash.herokuapp.com/api/"),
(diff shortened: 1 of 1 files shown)


sighphyre (approved):
Yeah this is an improvement, I agree with this change, since this will fall back to a GUID. We should probably make some docs about what this should be on our main docs at some point

thomasheartman:
Yeah, I agree. The `instanceId` is documented in some of the SDKs, but not as a general thing. Probably belongs on the SDK landing page, actually.

PR #126, opened 2023-02-22
add test to check if flexible rollout handles a null context correctly

### LINKED ISSUE #125: isEnabled for GradualRollout not working properly without a context
**Describe the bug**
When using the GradualRollout strategy with `isEnabled` method but not providing a context, the feature will always fall back to false, even if the configuration in Unleash is set to be enabled for 100% of users.

**To Reproduce**
1. Create a toggle with GradualRollout strategy and enable it for all customers. 
2. Use the SDK to query the toggle. E.g.:
```cs
if (_unleash.IsEnabled("show-welcome-message"))
{
    return Ok("Flag is activated!!!");
}

return Ok("Flag is not activated");
```

**Expected behavior**
We should receive `Ok("Flag is activated!!!")`

**Additional context**
If an empty context is provided, it will work as expected. 
It may be related with different stickiness values (has to be tested). But we suspect that this code: https://github.com/Unleash/unleash-client-dotnet/blob/60ac24c737f0907e072904c96a1e0af46bbda833/src/Unleash/Strategies/GradualRolloutUserIdStrategy.cs#L27-L29 and also https://github.com/Unleash/unleash-client-dotnet/blob/60ac24c737f0907e072904c96a1e0af46bbda833/src/Unleash/Strategies/GradualRolloutSessionIdStrategy.cs#L27-L29 requires a context with a valid parameter for the stickiness configuration to check the configuration of the toggle.

sighphyre:
This allows the flexible rollout strategy to correctly fallback to random when no context is passed. Closes #125

### DIFF
--- a/src/Unleash/Strategies/FlexibleRolloutStrategy.cs
+++ b/src/Unleash/Strategies/FlexibleRolloutStrategy.cs
@@ -68,7 +68,7 @@ namespace Unleash.Strategies
                         ?? context?.SessionId
                         ?? randomGenerator();
                 default:
-                    return context?.GetByName(stickiness) ?? randomGenerator();
+                    return context.GetByName(stickiness);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


gastonfournier (approved):
LGTM! I see you tested without the change and how the change fixed it. TDD FTW! :)

PR #127, opened 2023-02-28
feat: send sdk name during client registration

gardleopard:
This sets the sdk name as well as the version during the client registration.


gardleopard:
the old code did not make sense. It resolved to much more than we need and the name doesnt help identifying that it is the dotnet sdk. Hardcoded this instead.

what was fetched in the old code 
```
"Unleash.Client, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null:v1.0.0" 
   ```

sighphyre (approved):
G

PR #135, opened 2023-05-19
feat: error events

daveleek:
# Description

Adds a way to register callback handlers for errors that occur within the Unleash .NET SDK.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] Running existing unit tests
- [x] Adding more unit tests
- [ ] Will run in sample app

# Checklist:


src/Unleash/DefaultUnleash.cs
```diff
@@ -64,7 +65,7 @@ public DefaultUnleash(UnleashSettings settings, bool overrideDefaultStrategies,
             strategies = SelectStrategies(strategies, overrideDefaultStrategies);
             strategyMap = BuildStrategyMap(strategies);
 
-            services = new UnleashServices(settings, strategyMap);
+            services = new UnleashServices(settings, EventConfig, strategyMap);
```
sighphyre, line 68: Isn't this always null at this point?

src/Unleash/DefaultUnleash.cs
```diff
@@ -286,6 +287,7 @@ private void EmitImpressionEvent(string type, UnleashContext context, bool enabl
             catch (Exception ex)
             {
                 Logger.Error($"UNLEASH: Emitting impression event callback threw exception: {ex.Message}");
+                EventConfig?.RaiseError(new ErrorEvent() { Error = ex, ErrorType = ErrorType.ImpressionEvent });
```
sighphyre, line 290: This feels brave. What situations can cause the impression events to fail that wouldn't also impact the error events?

src/Unleash/Scheduling/FetchFeatureTogglesTask.cs
```diff
@@ -28,20 +31,32 @@ internal class FetchFeatureTogglesTask : IUnleashScheduledTask
...
             this.etagFile = etagFile;
         }
 
         public async Task ExecuteAsync(CancellationToken cancellationToken)
         {
-            var result = await apiClient.FetchToggles(Etag, cancellationToken).ConfigureAwait(false);
+            FetchTogglesResult result;
+            try
+            {
+                result = await apiClient.FetchToggles(Etag, cancellationToken).ConfigureAwait(false);
```
sighphyre, line 52: Looks like this call already raises an event, what does this catch block now do?

tests/Unleash.Tests/Internal/ErrorEvents_Tests.cs
```diff
@@ -0,0 +1,273 @@
...
+                ErrorEvent = evt => { callbackEvent = evt; }
+            };
+            var unleashClient = new UnleashApiClient(httpClient, new DynamicNewtonsoftJsonSerializer(), new UnleashApiClientRequestHeaders(), eventConfig: callbackConfig);
+            var cancellationTokenSource = new CancellationTokenSource();
+
+            // Act
+            var result = unleashClient.FetchToggles("123", cancellationTokenSource.Token).Result;
+
+            // Assert
+            callbackEvent.Should().NotBeNull();
```
sighphyre, line 48: I think it'd be nice if these tests actually checked that the response was correct as well. This should help us catch cases where we bubble stupid/incorrect events

sighphyre (commented):
I like the approach, looks sensible to me!

I would like to see an example for this so users can look at that rather than asking us how this works before I approve

src/Unleash/DefaultUnleash.cs
daveleek: Oooo good catch!

src/Unleash/DefaultUnleash.cs
daveleek: This is one of the things I was a bit uncertain of. But it can happen if I implement the impression event callback and then my code throws. Maybe not a good fit for the ErrorEvent as technically it's source is external to our code

src/Unleash/Scheduling/FetchFeatureTogglesTask.cs
daveleek: Catches very specific exceptions and let the rest bubble up as before

tests/Unleash.Tests/Internal/ErrorEvents_Tests.cs
daveleek: Ah yes thank you, unintentional omission. Will do

src/Unleash/Scheduling/FetchFeatureTogglesTask.cs
sighphyre: Ah, gotcha happens when http fails and the other event shouldn't be raised when that happens - sounds sane

sighphyre (approved):
Noice, I think this looks pretty good to me!

PR #147, opened 2023-06-13
fix: count variants

### LINKED ISSUE #138: Counting variants in the `ToggleCount` is incorrectly implemented
The following line of code in the `ToggleCount` class won't increase the variant counts in the `ConcurrentDictionary` because `Interlocked.Increment()` cannot be applied to properties.

https://github.com/Unleash/unleash-client-dotnet/blob/b5437c2c15e0eedfec788fe4273c16fc67b0e432/src/Unleash/Metrics/ToggleCount.cs#L31

There are some possible fixes, such as:

```csharp
// Using  StrongBox
 private ConcurrentDictionary<string, StrongBox<long>> variants = new ConcurrentDictionary<string, StrongBox<long>>();

Interlocked.Increment(ref variants[variantName].Value);
```

or

```csharp
// Using ConcurrentDictionary.AddOrUpdate
variants.AddOrUpdate(variantName, 1, (k, v) => Interlocked.Increment(ref v));
```

thomasheartman:
This PR adds a test and a fix for an issue we had when counting variants in the `ToggleCount` class. As reported in #138, the old way we counted didn't work.

I've kept the same testing style as applied in the other metrics bucket tests. Testing 100000 counts seems a little excessive, but it should at least prove that it works. Happy to change it if you think it's too much. 

Thanks to @nhathongly for the fix.

Fixes #138

## Type of change

# How Has This Been Tested?

- [x] Unit tests

**Test Configuration**:
* Firmware version: (I don't know. What do you want here?)
* Hardware: MacOS M1 chip
* Toolchain: dotnet version 7.0.302
* SDK: from `chore/make-dotnet-compatible` branch

# Checklist:

### DIFF
--- a/src/Unleash/Metrics/ToggleCount.cs
+++ b/src/Unleash/Metrics/ToggleCount.cs
@@ -27,7 +27,7 @@ namespace Unleash.Metrics
 
         public void Register(string variantName)
         {
-            variants.AddOrUpdate(variantName, 1, (k, v) => Interlocked.Increment(ref v));
+            variants.AddOrUpdate(variantName, 1, (k, v) => v + 1);
         }
 
         /// <summary>
--- a/src/Unleash/Metrics/ToggleCount.cs
+++ b/src/Unleash/Metrics/ToggleCount.cs
@@ -27,8 +27,7 @@ namespace Unleash.Metrics
 
         public void Register(string variantName)
         {
-            var current = variants.GetOrAdd(variantName, s => new long());
-            Interlocked.Increment(ref current);
+            variants.AddOrUpdate(variantName, 1, (k, v) => Interlocked.Increment(ref v));
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


daveleek (approved):
LGTM!

src/Unleash/Metrics/ToggleCount.cs
```diff
@@ -27,8 +27,7 @@ public void Register(bool active)
 
         public void Register(string variantName)
         {
-            var current = variants.GetOrAdd(variantName, s => new long());
-            Interlocked.Increment(ref current);
+            variants.AddOrUpdate(variantName, 1, (k, v) => Interlocked.Increment(ref v));
```
daveleek, line 30: I think we don't need Interlocked here. afaik a long is not a reference type - so it's not changed "in place" it's just copied and the new value is returned and set in a thread safe manner. I don't think it'll cause harm keeping it in though

sighphyre (approved):
Nice change, LGTM

src/Unleash/Metrics/ToggleCount.cs
sighphyre: Good eyes, agreed, it's noise here. It's not harmful to keep, outside of forcing an unnecessary lock, but it does nothing useful

thomasheartman: Ah, that's a good point! I'm happy to make that change. Should be updated in befbe3e05ee44e3be66a470ba46c760c1c239769

PR #151, opened 2023-07-13
chore: change context defaults to null - fixes flexiblestrat

daveleek:
# Description

This fixes an issue with Flexible rollout strategy - stickiness would be resolved to an empty (non null) string instead of defaulting to random number when context isn't provided

## Type of change

- [ ] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

- [x] Created new unit tests
- [x] Ran existing tests

# Checklist:

### DIFF
--- a/src/Unleash/DefaultUnleashContextProvider.cs
+++ b/src/Unleash/DefaultUnleashContextProvider.cs
@@ -8,9 +8,6 @@ namespace Unleash
         {
             Context = context ?? new UnleashContext
             {
-                UserId = string.Empty,
-                SessionId = string.Empty,
-                RemoteAddress = string.Empty,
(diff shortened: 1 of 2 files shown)


sighphyre (approved):
Ye!

thomasheartman (approved):
Nice! 😄 This is very nice 💯 How does it handle cases where the rollout isn't 100? Like 97? Is that done randomly then, or is it always false?

sighphyre:
> Nice! smile This is very nice 100 How does it handle cases where the rollout isn't 100? Like 97? Is that done randomly then, or is it always false?

Should fallback to the original calculation - so random

daveleek:
Yes, there's a random calculation and compared with stickiness

Issue #152, opened 2023-08-07
Unhandled TaskCanceledException by the FetchFeatureTogglesTask


rafaelpadovezi:
**Describe the bug**
Not sure if is a bug or the intended behavior. If the call to features API throws a timeout error (`TaskCanceledException`) the unleash is not initialized, considering a synchronous initialization. The [current code](https://github.com/Unleash/unleash-client-dotnet/blob/8158ccf9485e2433896609deff5faf2cb753de8d/src/Unleash/Scheduling/FetchFeatureTogglesTask.cs#L54) catches the `HttpRequestException` but do not catch the TaskCanceledException ([thrown by `SendAsync` when a timeout occurs](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.sendasync?view=net-7.0#system-net-http-httpclient-sendasync(system-net-http-httprequestmessage))).

**To Reproduce**

```cs
var settings = new UnleashSettings
{
    UnleashApi = new Uri("http://10.255.255.1/"), // just an address to force the timeout exception
};
var unleash = new UnleashClientFactory().CreateClient(settings, synchronousInitialization:true);
```

**Expected behavior**
The timeout exception should be handled, and the UnleashClient created.

**Desktop (please complete the following information):**
 - OS: Ubuntu 22.04
 - Browser [e.g. chrome, safari]: N/A

daveleek:
Thank you for raising this @rafaelpadovezi, we'll have a look and get back to you!

daveleek:
Hi again @rafaelpadovezi. We've had a look and compared it with other SDKs and how it is intended to work.
Synchronous initialization is a blocking operation and should either block execution until features have been fetched and initialized, block until the request times out and then crash, or crash because the server is unavailable/server errors etc.

So it's intended behavior. Catching `HttpRequestException` was done to implement Error events however, and will be amended to work as intended in the synchronous initialization operation (it should crash).

Setting synchronous initialization to `false` should allow the behavior I think you are asking for - it will schedule a fetch but initialize the Unleash client and make that immediately available. Feature evaluations will resolve to `false` until features have been fetched and properly initialized. To address/work with this there are a few options:

###  [Local backups](https://github.com/Unleash/unleash-client-dotnet#local-backup)
Unleash will update a local backup every time it has fetched an updated set of features. This backup is attempted loaded during initialization of Unleash.

[...]

PR #154, opened 2023-08-14
fix: catch exceptions during sync startup and return client

daveleek:
# Description

Fixes an issue during synchronous startup where exceptions like TaskCanceledException would crash everything instead of logging and just returning the client

Fixes # (issue)

#152 

## Type of change

# How Has This Been Tested?

- By running a test-application locally

# Checklist:


sighphyre (commented):
This looks subtle enough to deserve a test I think. Does the client that's returned when this errors actually do useful things? Is it supposed to have a bootstrap at this point (I wanna say yes)? Might be worth codifying this behaviour in a test so the silly old behaviour doesn't come back, WDYT?

PR #156, opened 2023-08-15
fix: impression data should now contain static appName even if its not set

sighphyre:
This patches an issue in the SDK where the appName was not being set correctly when applying impression data. This was caused by passing the base context, rather than the enriched context


daveleek (approved):
LGTM!

PR #168, opened 2023-10-16
feat: dependent feature toggles

daveleek:
# Description

Adds support for dependent features in .NET SDK. Also adds a warn once component

## Type of change

# How Has This Been Tested?

- [x] Running spec tests locally
- [x] Adding and running a suite of unit tests

# Checklist:

### DIFF
--- a/src/Unleash/DefaultUnleash.cs
+++ b/src/Unleash/DefaultUnleash.cs
@@ -105,7 +105,10 @@ namespace Unleash
 
         public bool IsEnabled(string toggleName, UnleashContext context, bool defaultSetting)
         {
-            return CheckIsEnabled(toggleName, context, defaultSetting).Enabled;
+            var enabled = CheckIsEnabled(toggleName, context, defaultSetting).Enabled;
+            RegisterCount(toggleName, enabled);
+
+            return enabled;
         }
 
         private FeatureEvaluationResult CheckIsEnabled(
@@ -118,8 +121,6 @@ namespace Unleash
             var enhancedContext = context.ApplyStaticFields(settings);
             var enabled = DetermineIsEnabledAndStrategy(toggleName, featureToggle, enhancedContext, defaultSetting, out var strategy);
             var variant = DetermineVariant(enabled, featureToggle, strategy, enhancedContext, defaultVariant);
-
-            RegisterCount(toggleName, enabled);
 
             if (featureToggle?.ImpressionData ?? false)
             {
@@ -193,10 +194,10 @@ namespace Unleash
                 {
                     return dependency.Variants.Contains(GetVariant(dependency.Feature, context).Name);
                 }
-                return IsEnabled(dependency.Feature, context, false);
+                return CheckIsEnabled(dependency.Feature, context, false).Enabled;
             }
 
-            return !IsEnabled(dependency.Feature, context, false);
+            return !CheckIsEnabled(dependency.Feature, context, false).Enabled;
         }
 
         private Variant DetermineVariant(bool enabled,
@@ -244,6 +245,8 @@ namespace Unleash
             var toggle = GetToggle(toggleName);
 
             var evaluationResult = CheckIsEnabled(toggleName, context, false, defaultValue);
+
+            RegisterCount(toggleName, evaluationResult.Enabled);
 
             RegisterVariant(toggleName, evaluationResult.Variant);
 
--- a/src/Unleash/DefaultUnleash.cs
+++ b/src/Unleash/DefaultUnleash.cs
@@ -173,15 +173,15 @@ namespace Unleash
 
         private bool ParentDependenciesAreSatisfied(FeatureToggle featureToggle, UnleashContext context)
         {
-            return featureToggle.Dependencies.All(d => DependenciesSatisfied(d, context));
+            return featureToggle.Dependencies.All(d => DependenciesSatisfied(featureToggle, d, context));
         }
 
-        private bool DependenciesSatisfied(Dependency dependency, UnleashContext context)
+        private bool DependenciesSatisfied(FeatureToggle featureToggle, Dependency dependency, UnleashContext context)
         {
             var parentToggle = GetToggle(dependency.Feature);
             if (parentToggle == null)
             {
-                warnOnce.Warn(dependency.Feature, $"UNLEASH: Parent feature toggle {dependency.Feature} was not found in the cache, the evaluation of this dependency will always be false");
+                warnOnce.Warn(dependency.Feature + featureToggle.Name, $"UNLEASH: Parent feature toggle {dependency.Feature} was not found in the cache, the evaluation of this dependency will always be false");
                 return false;
             }
 
--- a/src/Unleash/DefaultUnleash.cs
+++ b/src/Unleash/DefaultUnleash.cs
@@ -8,6 +8,7 @@ namespace Unleash
     using System.Linq;
     using System.Threading;
     using Unleash.Events;
+    using Unleash.Utilities;
     using Unleash.Variants;
 
     /// <inheritdoc />
@@ -38,6 +39,8 @@ namespace Unleash
 
         internal readonly UnleashServices services;
 
+        private readonly WarnOnce warnOnce;
+
         ///// <summary>
         ///// Initializes a new instance of Unleash client with a set of default strategies.
         ///// </summary>
@@ -58,6 +61,8 @@ namespace Unleash
             var currentInstanceNo = Interlocked.Increment(ref InitializedInstanceCount);
 
             this.settings = settings;
+
+            warnOnce = new WarnOnce(Logger);
 
             var settingsValidator = new UnleashSettingsValidator();
             settingsValidator.Validate(settings);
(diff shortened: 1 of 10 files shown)


kwasniew (commented):
Overall it looks very good. Just some questions:
* why do we have to write all the tests inside this repo. Isn't the specification covering everything that we need?
* can we add a test that shows that metrics are not called on the dependent features
* do we have impression events in .net SDK? If so, how do they behave on the dependent features?

daveleek:
Good questions @kwasniew!

> why do we have to write all the tests inside this repo. Isn't the specification covering everything that we need?

it's been convention so far to also add local unit tests, helps with development and debugging etc, I'll rope @sighphyre in for opinions though

> can we add a test that shows that metrics are not called on the dependent features

Coming right up!

> do we have impression events in .net SDK? If so, how do they behave on the dependent features?

src/Unleash/DefaultUnleash.cs
```diff
@@ -155,9 +160,43 @@ public bool IsEnabled(string toggleName, UnleashContext context, bool defaultSet
...
+        {
+            return featureToggle.Dependencies.All(d => DependenciesSatisfied(d, context));
+        }
+
+        private bool DependenciesSatisfied(Dependency dependency, UnleashContext context)
+        {
+            var parentToggle = GetToggle(dependency.Feature);
+            if (parentToggle == null)
+            {
+                warnOnce.Warn(dependency.Feature, $"UNLEASH: Parent feature toggle {dependency.Feature} was not found in the cache, the evaluation of this dependency will always be false");
```
sighphyre, line 183: This will log once per parent that's disabled. So if you have toggle A and toggle B that both depend on toggle C, which is missing, once A is evaluated and this message shows,  the B will never get a warning message. Is that what we want? (Genuinely asking, I've heard about warn once but this is the first time I've seen it)

src/Unleash/Internal/Dependency.cs
```diff
@@ -0,0 +1,24 @@
+namespace Unleash.Internal {
+    public class Dependency {
```
sighphyre, line 2: Does this need to be public? Totally okay if it has to be or it makes sense to be but I'd like to hide stuff we don't want other people to permanently depend on - stupidly hard to undo public dependencies

src/Unleash/Utilities/WarnOnce.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Unleash.Utilities
+{
+	internal class WarnOnce
+	{
+        private readonly ILog logger;
+        private readonly HashSet<string> seen = new HashSet<string>();
+
+        public WarnOnce(ILog logger)
+		{
```
sighphyre, line 13: Nit: Can we give this guy a little bit of autoformat love before we merge?

tests/Unleash.Tests/Integration/ClientSpecificationTests.cs
```diff
@@ -38,7 +38,7 @@ static TestFactory()
 
             using (var client = new HttpClient())
             {
-                var csTestsVersion = "v4.3.3";
+                var csTestsVersion = "v4.5.1";
```
sighphyre, line 41: 4.5.2 is out! Adds a test case, which is relevant to dependent features!

```suggestion
                var csTestsVersion = "v4.5.2";
```

sighphyre:
> Good questions @kwasniew!
> 
> > why do we have to write all the tests inside this repo. Isn't the specification covering everything that we need?
> 
> it's been convention so far to also add local unit tests, helps with development and debugging etc, I'll rope @sighphyre in for opinions though


I'd lean towards reducing the tests here. This is a really well covered feature in the client specs, I don't see much in your tests that aren't doing what the spec is doing. Usually I'm fussy about having tests that cover cases the spec doesn't for your particular implementation or in cases where a unit test is much easier to deal with than the spec. This is a pretty simple feature though

Things I'd like to see tested:

- What does metrics do here? (don't think this is covered, we could use some tests here)
- Does the response parse correctly and set the right defaults (this does look covered)
- Does the feature work as expected (covered through the client spec, I don't see anything you've done in the code that would require additional testing for the actual implementation)

src/Unleash/Internal/Dependency.cs
daveleek: It's a property of public class `FeatureToggle` so needs to be as accessible as it's containing object

src/Unleash/DefaultUnleash.cs
daveleek: https://github.com/Unleash/unleash-client-node/pull/520

In the intro text:
> warning events for missing dependencies are reported once
> `this.warnedDependencies[missingDependency + name] = true;`

So no, should warn once per child and dependency it seems, I'll fix

daveleek: I've addressed this now

daveleek:
@sighphyre 

> I'd lean towards reducing the tests here...  

> ... or in cases where a unit test is much easier to deal with than the spec. This is a pretty simple feature though

I guess I could trim some tests, but they've been valuable to verify understanding at development time and as debug tools so I'd like to keep a few in

> Things I'd like to see tested:
> What does metrics do here? (don't think this is covered, we could use some tests here)

Added a test for that

> What should impression events do (think you're working on this already)

Also added a test for this

> Does the response parse correctly and set the right defaults (this does look covered)

This should be covered, I have attempted to at least

>Does the feature work as expected (covered through the client spec, I don't see anything you've done in the code that would require additional testing for the actual implementation)

👍

src/Unleash/Internal/Dependency.cs
sighphyre: Okay, we chatted offline, it'd be great but this is much harder than simply this right now. I'm okay with this

sighphyre:
> @sighphyre
> 
> > I'd lean towards reducing the tests here...
> 
> > ... or in cases where a unit test is much easier to deal with than the spec. This is a pretty simple feature though
> 
> I guess I could trim some tests, but they've been valuable to verify understanding at development time and as debug tools so I'd like to keep a few in

If you find them useful I'm happy for them to stay. At some point we should get to the point where it's nice to use the client spec but that's definitely out of scope here (JS and Ruby are pretty good at this)

sighphyre (approved):
Happy if you are

tests/Unleash.Tests/Serialization/DynamicJsonSerializerTests.cs
```diff
@@ -136,5 +136,19 @@ public void Serializes_ImpressionData_Property()
...
+        [Test]
+        public void Dependent_Feature_Enabled_Defaults_To_True() {
+            // Arrange
+            var path = Path.Combine(TestContext.CurrentContext.TestDirectory, "App_Data", "dependent-features-missing-enabled.json");
+            var originalJson = File.ReadAllText(path);
+
+            // Act
+            var deserialized = JsonConvert.DeserializeObject<ToggleCollection>(originalJson);
+            var toggle = deserialized.Features.First(f => f.Name == "enabled-child");
+            toggle.Should().NotBeNull();
```
kwasniew, line 149: missing //assert?

tests/Unleash.Tests/Internal/Dependent_Features_Tests.cs
```diff
@@ -0,0 +1,610 @@
...
+            state.Version = 2;
+            var unleash = CreateUnleash(appname, state);
+
+            // Act
+            var result = unleash.IsEnabled("child-1");
+
+            // Assert
+            unleash.services.MetricsBucket.StopCollectingMetrics(out var bucket);
+            var childMetrics = bucket.Toggles.Single(t => t.Key == "child-1").Value;
+            (childMetrics.No + childMetrics.Yes).Should().Be(1L);
```
kwasniew, line 129: personally I prefer explicit numbers in tests instead of calculations

kwasniew (approved):
I'm curious if you used client spec in development? You said that manually written tests were useful in development, but in Node SDK I used client spec for that. On the other hand I had to wade through hundreds lines of test code to find the metrics test I wanted to check. I'm trying to understand what exactly are we missing from the shared client spec tests. Are the .net tests easier to work with?

tests/Unleash.Tests/Serialization/DynamicJsonSerializerTests.cs
daveleek: Good catch!

tests/Unleash.Tests/Internal/Dependent_Features_Tests.cs
daveleek: Fair point, the explicit number was in the .Be part of the assertion - but I'll add a separate line for each to not sum the metrics no and yes!

daveleek:
> I'm curious if you used client spec in development?

I did use that as well, but mostly after being done with large portions of code. And whenever something was wrong I could repro it with a specific unit test and debug from there.

> I'm trying to understand what exactly are we missing from the shared client spec tests. Are the .net tests easier to work with?

I'm not going to say the .NET tests are awesome, but for me it boils down a combination of two things:
- Habit - I'm used to unit tests showing me what I'm building and my understanding of it is sane)
- Convenience - Unit tests are (for now) easier to work with on a case by case basis. Can probably instrument the spec test helpers to give more granular control over tests to run, but I have no idea of the amount of work involved so short term easier to just keep doing what I'm doing

I'm also less comfortable around the implementation of the spec parsing and knowing what might be missing there but still providing false positives - so I use Unit tests as a way of reassuring myself

With that said I find the spec massively useful as an overview into where I'm at with the implementation 30/40/90% "done" / green board check off, and also as a source of knowledge/insight into the feature and how it's expected to work

PR #170, opened 2023-10-27
feat: log errors instead of warnings if instance count is over 10

daveleek:
# Description

Starts logging errors (instead of warnings) when unleash instance count is higher than 10


ivarconr (approved):

sighphyre (approved):
Nice

PR #171, opened 2023-10-30
fix: add a seed to variant utils murmur hashing to fix variant distri…

daveleek:
**What**

Uses a new seed for ensuring a fair distrbution for variants.

**Background**

After a customer reported that variant distribution seemed skewed we performed some testing and found that since we use the same hash string for both gradual rollout and variant allocation we'd reduced the set of groups we could get to whatever percentage our gradual rollout was set.

**Example**

Take a gradualRollout of 10%, this will select normalized hashes between 1 and 10, when we then again hash the same string that gave us between 1 and 10, but with modulo 1000 for variants, this will only give us 100 possible groups, instead of the expected 1000.

**Fix**

Force the normalization to accept a seed, and make sure to use a new seed when normalizing the variant distribution hash.

**Worth noting**

This will require release 4.0.0, since we're changing the signature of public methods.

### DIFF
--- a/src/Unleash/Strategies/StrategyUtils.cs
+++ b/src/Unleash/Strategies/StrategyUtils.cs
@@ -12,14 +12,14 @@ namespace Unleash.Strategies
         /// <summary>
         /// Takes to string inputs concat them, produce a hashCode and return a normalized value between 0 and 100;
         /// </summary>
-        public static int GetNormalizedNumber(string identifier, string groupId, int normalizer = 100)
+        public static int GetNormalizedNumber(string identifier, string groupId, uint randomSeed, int normalizer = 100)
         {
             const int one = 1;
             const string separator = ":";
 
             byte[] bytes = Encoding.UTF8.GetBytes(string.Concat(groupId, separator, identifier));
 
-            using (var algorithm = MurmurHash.Create32())
+            using (var algorithm = MurmurHash.Create32(randomSeed))
             {
                 var hash = algorithm.ComputeHash(bytes);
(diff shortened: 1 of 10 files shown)


src/Unleash/Strategies/StrategyUtils.cs
```diff
@@ -12,14 +12,14 @@ internal class StrategyUtils
         /// <summary>
         /// Takes to string inputs concat them, produce a hashCode and return a normalized value between 0 and 100;
         /// </summary>
-        public static int GetNormalizedNumber(string identifier, string groupId, int normalizer = 100)
+        public static int GetNormalizedNumber(string identifier, string groupId, uint randomSeed, int normalizer = 100)
```
gastonfournier, line 15:
```suggestion
        public static int GetNormalizedNumber(string identifier, string groupId, uint randomSeed = 0, int normalizer = 100)
```
Can you just default to zero?

gastonfournier (approved):
LGTM, defaulting to zero is more of a cosmetic but might be a nice addition

src/Unleash/Strategies/StrategyUtils.cs
daveleek: I went with what Chris implemented in the Java client. Open for discussion, but forcing a choice was intentional:
https://github.com/Unleash/unleash-client-java/pull/220

gastonfournier: Good, I considered that possibility

PR #174, opened 2023-11-14
fix: make segment ids deserialize as ints instead of strings

### LINKED ISSUE #173: System.Text.Json.JsonException when dezerializing json to Unleash.Internal.ToggleCollection
**Describe the bug**
The id in Segment is defined as a string but unleash seems to return a json number. This causes a System.Text.Json.JsonException: The JSON value could not be converted to Unleash.Internal.Segment when using System.Text.Json for dezerialization. The bug is not noticable when using Newtonsoft.

unleash-client-dotnet version: 4.0.0

Segment.cs: https://github.com/Unleash/unleash-client-dotnet/blob/cccf3efe3cb2effd5580834efc864b1f0542160e/src/Unleash/Internal/Segment.cs
```
﻿using System;
using System.Collections.Generic;
using System.Text;

namespace Unleash.Internal
{
    public class Segment
    {
        public string Id { get; }
        public List<Constraint> Constraints { get; }
        public Segment(string id, List<Constraint> constraints = null)
        {
            Id = id;
            Constraints = constraints ?? new List<Constraint>();
        }
    }
}

```
**To Reproduce**
1. Create a simple app that reference "unleash.client" Version="4.0.0".
2. Add  JsonSerializer = new TextJsonSerializer() to UnleashSettings.
3. Wrap System.Text.Json serializer:

```
internal class TextJsonSerializer : IJsonSerializer
{

    T IJsonSerializer.Deserialize<T>(Stream stream)
    {
        var options = new JsonSerializerOptions
        {
            PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
        };
        var instance = JsonSerializer.Deserialize<T>(stream, options);
        if (instance == null)
        {
            throw new Exception("The result of serializing feature toggle is null");
        }

        return instance;
    }


    void IJsonSerializer.Serialize<T>(Stream stream, T instance)
    {
        JsonSerializer.Serialize<T>(stream, instance);
    }
}
```
4. Fetch feature toggles from unleash
5. See the error when dezerializing.

**Expected behavior**
instance should be Unleash.Internal.ToggleCollection after dezerialization.

**Suggested fix**
Change type of id in Segment.cs to int.

daveleek:
# Description

Segment ids were being sent as numbers but deserialized into strings, which can cause deserialization errors. This fix changes the local type of segment ids to be ints

Fixes #173 

## Type of change

# How Has This Been Tested?

- [x] Running all unit tests

# Checklist:

### DIFF
--- a/src/Unleash/Internal/ActivationStrategy.cs
+++ b/src/Unleash/Internal/ActivationStrategy.cs
@@ -8,15 +8,15 @@ namespace Unleash.Internal
         public string Name { get; }
         public Dictionary<string, string> Parameters { get; }
         public List<Constraint> Constraints { get; }
-        public List<string> Segments { get; }
+        public List<int> Segments { get; }
         public List<VariantDefinition> Variants { get; }
 
-        public ActivationStrategy(string name, Dictionary<string, string> parameters, List<Constraint> constraints = null, List<string> segments = null, List<VariantDefinition> variants = null)
+        public ActivationStrategy(string name, Dictionary<string, string> parameters, List<Constraint> constraints = null, List<int> segments = null, List<VariantDefinition> variants = null)
         {
             Name = name;
             Parameters = parameters ?? new Dictionary<string, string>();
             Constraints = constraints ?? new List<Constraint>();
-            Segments = segments ?? new List<string>();
+            Segments = segments ?? new List<int>();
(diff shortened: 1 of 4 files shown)


sighphyre (approved):
G

PR #179, opened 2023-11-30
feat: core switchable parallel yggdrasil engine integration

daveleek:
# Description

This is a port of the core parts of `feat/yggdrasil` made parallel/switchable, so we can have either engine depending on choices made in UnleashSettings.

All current tests are passing as the engine defaults to off. will port tests that test the integration properly.
Build should currently fail due to missing nuget dependency Yggdrasil.Engine.

I've introduced a base-branch for Yggdrasil development that this PR is aimed at - named `yggdrasil`

### DIFF
--- a/src/Unleash/DefaultUnleash.cs
+++ b/src/Unleash/DefaultUnleash.cs
@@ -57,6 +57,17 @@ namespace Unleash
         ///// <param name="overrideDefaultStrategies">When true, it overrides the default strategies.</param>
         ///// <param name="strategies">Custom strategies.</param>
         public DefaultUnleash(UnleashSettings settings, bool overrideDefaultStrategies, params IStrategy[] strategies)
+            : this(settings, overrideDefaultStrategies, strategies, yggdrasilStrategies: null)
+        { }
+
+        ///// <summary>
+        ///// Initializes a new instance of Unleash client.
+        ///// </summary>
+        ///// <param name="config">Unleash settings</param>
+        ///// <param name="overrideDefaultStrategies">When true, it overrides the default strategies.</param>
+        ///// <param name="strategies">Custom strategies.</param>
+        ///// <param name="yggdrasilStrategies">Yggdrasil custom strategies.</param>
+        public DefaultUnleash(UnleashSettings settings, bool overrideDefaultStrategies, IStrategy[] strategies, params Yggdrasil.IStrategy[] yggdrasilStrategies)
         {
             var currentInstanceNo = Interlocked.Increment(ref InitializedInstanceCount);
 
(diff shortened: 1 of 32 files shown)


src/Unleash/Unleash.csproj
```diff
@@ -1,13 +1,16 @@
...
+  <!--
   <PropertyGroup Condition=" '$(OS)' == 'Windows_NT' ">
     <TargetFrameworks>netstandard2.0;net48;net472;net47;net461;net46;net451;net45</TargetFrameworks>
   </PropertyGroup>
 
   <PropertyGroup Condition=" '$(OS)' != 'Windows_NT' ">
     <TargetFrameworks>netstandard2.0</TargetFrameworks>
+  </PropertyGroup>-->
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
```
daveleek, line 12: This needs changing in Yggdrasil.Engine first before we can go back to shipping more than just .NET 6 binaries

src/Unleash/UnleashContext.cs
```diff
@@ -6,16 +6,8 @@ namespace Unleash
     /// <summary>
     /// A context which the feature request should be validated against. Usually scoped to a web request through an implementation of IUnleashContextProvider.
     /// </summary>
-    public class UnleashContext
+    public class UnleashContext : Yggdrasil.Context
```
daveleek, line 9: This made things easier to work with. But not sure if I'm a fan of introducing object model reliance on Yggdrasil.

src/Unleash/Internal/Variant.cs
```diff
@@ -18,5 +19,13 @@ public Variant(string name, Payload payload, bool enabled, bool feature_enabled)
         public Payload Payload { get; }
         public bool IsEnabled { get; }
         public bool FeatureEnabled { get; internal set; }
+
+        public static Variant FromEngineVariant(YggdrasilVariant variant)
+        {
+            if (variant == null)
+                return DISABLED_VARIANT;
+
+            return new Variant(variant.Name, new Payload(variant.Payload.PayloadType, variant.Payload.Value), variant.Enabled, variant.FeatureEnabled);
```
daveleek, line 28: Maybe I should just move this back out, it's only being used one place anyways

src/Unleash/DefaultUnleash.cs
```diff
@@ -57,6 +57,17 @@ public DefaultUnleash(UnleashSettings settings, params IStrategy[] strategies)
...
+        { }
+
+        ///// <summary>
+        ///// Initializes a new instance of Unleash client.
+        ///// </summary>
+        ///// <param name="config">Unleash settings</param>
+        ///// <param name="overrideDefaultStrategies">When true, it overrides the default strategies.</param>
+        ///// <param name="strategies">Custom strategies.</param>
+        ///// <param name="yggdrasilStrategies">Yggdrasil custom strategies.</param>
+        public DefaultUnleash(UnleashSettings settings, bool overrideDefaultStrategies, IStrategy[] strategies, params Yggdrasil.IStrategy[] yggdrasilStrategies)
```
daveleek, line 70: Decided on adding an overload for Yggdrasil strategies. They share name and some of the APIs are very much alike, but I don't like introducing inheritance between these two as that makes a mess of things. So for now Custom strategies for yggdrasil is a consciously made opt-in and shouldn't break things for existing users

PR #181, opened 2023-12-07
fix: use stringcontent for metrics and register so we can set headers

daveleek:
# Description

Changes requests to metrics and register in UnleashApiClient, now uses StringContent instead of StreamContent so we can set Content-Length header properly. Makes the SDK play nicer together with Edge

Fixes https://github.com/Unleash/unleash-edge/issues/357

## Type of change

# How Has This Been Tested?

Running locally against Edge

# Checklist:


chriswk (approved):
Nice. Thanks for this.

PR #196, opened 2024-01-09
fix: IN and NOT_IN Inverted now actually inverts the result

sjaanus:
When using the IN and NOT_IN operators with the 'inverted' option turned on, the 'inverted' flag was previously ignored. This PR addresses this issue by ensuring that when 'inverted' is applied, it effectively inverts the result.


daveleek (approved):
Nice one! LG!

PR #200, opened 2024-01-19
fix: make semver logging actually log context value that failed parsing

sighphyre:
This patches the logging in the semver constraint to actually take the parameter that it needs to log, rather than throwing around a format string with nothing to format into it. This would previously cause some log implementations to crash and others to log a very unhelpful error message.


daveleek (approved):
Yeah that should improve things

PR #203, opened 2024-01-22
chore: update license in csproj to use PackageLicenseExpression

sighphyre:
Updates the license declaration to use PackageLicenseExpression rather than PackageLicenseUrl, since that's been deprecated


daveleek (approved):
LG!

PR #213, opened 2024-04-16
fix: throw when fetch features fails

### LINKED ISSUE #210: synchronousInitialization: true should throw exception if exception happens in the init flow (network, authorization, etc)
[2-2009](https://linear.app/unleash/issue/2-2009/net-sdk-unhandled-background-exceptions)

Tracking on EG board

Needs comms from SR. @david  to shout when there's needs

daveleek:
# Description

Failing fetches should throw during synchronous initializations.
Introduces a way to get this done that should be forwards and backwards compatible, but has a small public signature change in the ApiClient used internally.

Fixes #210 

## Type of change

# How Has This Been Tested?

- Added unit tests
- Ran existing tests

# Checklist:

### DIFF
--- a/src/Unleash/Communication/IUnleashApiClient.cs
+++ b/src/Unleash/Communication/IUnleashApiClient.cs
@@ -6,7 +6,7 @@ namespace Unleash.Communication
 {
     internal interface IUnleashApiClient
     {
-        Task<FetchTogglesResult> FetchToggles(string etag, CancellationToken cancellationToken);
+        Task<FetchTogglesResult> FetchToggles(string etag, CancellationToken cancellationToken, bool throwOnFail = false);
         Task<bool> RegisterClient(ClientRegistration registration, CancellationToken cancellationToken);
         Task<bool> SendMetrics(ThreadSafeMetricsBucket metrics, CancellationToken cancellationToken);
     }
--- a/src/Unleash/Communication/UnleashApiClient.cs
+++ b/src/Unleash/Communication/UnleashApiClient.cs
@@ -58,7 +58,7 @@ namespace Unleash.Communication
             this.projectId = projectId;
         }
 
-        public async Task<FetchTogglesResult> FetchToggles(string etag, CancellationToken cancellationToken)
+        public async Task<FetchTogglesResult> FetchToggles(string etag, CancellationToken cancellationToken, bool throwOnFail = false)
         {
             if (featureRequestsToSkip > featureRequestsSkipped)
             {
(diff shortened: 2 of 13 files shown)


tests/Unleash.Tests/MockedUnleashSettings.cs
```diff
@@ -8,10 +8,10 @@ namespace Unleash.Tests
 {
     public class MockedUnleashSettings : UnleashSettings
     {
-        public MockedUnleashSettings(bool mockFileSystem = true)
+        public MockedUnleashSettings(bool mockFileSystem = true, string instanceTag = "test instance 1")
         {
             AppName = "test";
-            InstanceTag = "test instance 1";
+            InstanceTag = instanceTag;
```
daveleek, line 14: This keeps failing when used in multiple test files being run simultaneously

src/Unleash/Scheduling/FetchFeatureTogglesTask.cs
```diff
@@ -58,6 +61,8 @@ public async Task ExecuteAsync(CancellationToken cancellationToken)
                 throw new UnleashException("Exception while fetching from API", ex);
             }
 
+            ready = true;
```
daveleek, line 64: This might have to get moved if we add support for the ready event

src/Unleash/Communication/IUnleashApiClient.cs
```diff
@@ -6,7 +6,7 @@ namespace Unleash.Communication
 {
     internal interface IUnleashApiClient
     {
-        Task<FetchTogglesResult> FetchToggles(string etag, CancellationToken cancellationToken);
+        Task<FetchTogglesResult> FetchToggles(string etag, CancellationToken cancellationToken, bool throwOnFail = false);
```
daveleek, line 9: public method but on an internal interface. this should be ok?

sighphyre: Yeah I think so

src/Unleash/Communication/UnleashApiClient.cs
```diff
@@ -71,7 +71,7 @@ public async Task<FetchTogglesResult> FetchToggles(string etag, CancellationToke
             }
 
             featureRequestsSkipped = 0;
-
+
```
sighphyre, line 74: Not a blocker, but formatter doesn't sort this out

src/Unleash/Scheduling/FetchFeatureTogglesTask.cs
sighphyre: Smells like a later problem to me

tests/Unleash.Tests/MockedUnleashSettings.cs
sighphyre: That seems like a big change though. Any idea why?

sighphyre: Okay we chatted, sounds like this was a legacy problem

sighphyre (approved):
LGTM

PR #216, opened 2024-04-19
:bug: Fix empty user Id return false from IsEnabled #215

### LINKED ISSUE #215: Empty userId breaks gradual rollout 100% expected behavior.
**Describe the bug**
When passing an empty userId (string.Empty) the toggle is not enabled. Although the toggle have a gradual rollout strategy of 100%. 
When passing a null or a non-empty value, the same toogle is enabled as expected. 

**To Reproduce**
Create a toggle with a gradual rollout strategy.
Pass the userId as string.empty ("") : 
```csharp
  var unleashEnabled = unleash.IsEnabled("togglename", new UnleashContext()
  {
      
      UserId = string.Empty

  });
```

**Expected behavior**
The value of unleashedEnabled should be always true, since the gradual rollout is 100% and there is not constraints. 
**Screenshots**

**Version ** 
```xml
    <PackageReference Include="Unleash.Client" Version="4.1.8" />

```

**Additional context**
Did my own digging and I think that the source is the calculation of `stickinessId` https://github.com/Unleash/unleash-client-dotnet/blob/main/src/Unleash/Strategies/FlexibleRolloutStrategy.cs#L43
When userId is empty, it is returned as the first choice of stickinessId but then we do the `if (!string.IsNullOrEmpty(stickinessId))
` which will be false, always.

karimkod:
# Description
Allow stickiness to be an empty string.

Fixes #215

### DIFF
--- a/src/Unleash/Strategies/FlexibleRolloutStrategy.cs
+++ b/src/Unleash/Strategies/FlexibleRolloutStrategy.cs
@@ -40,7 +40,7 @@ namespace Unleash.Strategies
                 groupId = "";
             }
 
-            if (!string.IsNullOrEmpty(stickinessId))
+            if (!(stickinessId is null))
             {
                 var normalizedUserId = StrategyUtils.GetNormalizedNumber(stickinessId, groupId, 0);
                 return percentage > 0 && normalizedUserId <= percentage;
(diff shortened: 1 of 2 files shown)


sighphyre (approved):
Thank you @karimkod, that looks great!

karimkod:
Just for my knowledge, when will this be released? So we can upgrade from our side as well.
Thanks 😊

andreas-unleash:
Just released v4.1.9 @karimkod 🚀

karimkod:
Thanks!!!

PR #217, opened 2024-04-29
chore:dev improvements

daveleek:
# Description

## Removes the solution dependency on the Unleash.Tests.Benchmarks project
- Should now not throw warnings or errors while building the solution with .NET 6++
- Project still there, just not in use in the SLN

## README Documentation update
- Adds some documentation to the README on how development works and requirements
- Adds documentation on doing releases


sighphyre (approved):
Noice!

PR #220, opened 2024-05-07
Fixed typo with types in Readme

alvinometric:
# Description

I run into a build error if I don't specify this. It's specified in other examples so I think it's probably a typo?


daveleek (approved):
Nice one, good catch!

PR #224, opened 2024-06-26
chore: yggdrasil

nunogois:
https://linear.app/unleash/issue/2-1545/yggdrasil-net-integrate-into-unleashclient-net-sdk

This incorporates [Yggdrasil](https://github.com/Unleash/yggdrasil) into the .NET SDK, using the new [Unleash.Yggdrasil](https://www.nuget.org/packages/Unleash.Yggdrasil) NuGet package.

The PR tries to target as small of a change surface as possible, however some things simply no longer made sense or needed to change to accommodate this change to the new Yggdrasil engine.

After this change we can consider removing a big part of what can now be considered "dead code". There are a lot of refactoring opportunities. We should also consider favoring state literal strings in tests over constructing the state with the current classes. However all of this is out of scope for this PR. 

Some changes that affect the public interface and usability of the SDK include:

 - Due to [Unleash.Yggdrasil](https://www.nuget.org/packages/Unleash.Yggdrasil)'s target of [.NET Standard 2.0](https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0), we're dropping support for:
   - .NET Framework 4.7;
   - .NET Framework 4.6.1;
   - .NET Framework 4.6;
   - .NET Framework 4.5.1;
   - .NET Framework 4.5;
 - In the Unleash client, we no longer expose:
   - `FeatureToggles`;
   - `GetVariants`;
 - `IStrategy` -> `Yggdrasil.IStrategy` at Unleash client creation, for custom strategies;
 - `IToggleBootstrapProvider.Read()` should now return a state string instead of a `ToggleCollection`;
 - `UnleashServices` now logs an error and emits a `FileCache` error whenever `YggdrasilEngine` fails to load initial state;

Some changes that are internal to the SDK include:

 - `FetchTogglesResult` drops `ToggleCollection` in favor of `State`;
 - `SendMetrics` now takes in an `Yggdrasil.MetricsBucket` instead of a `ThreadSafeMetricsBucket`;
 - `CachedFilesLoader` no longer needs a serializer, since it uses the file string content as is;
 - `CachedFilesLoader` drops `InitialToggleCollection` in favor of `InitialState`;
 - `UnleashServices` now holds a list of `DefaultStrategyNames` in order to register them as part of metrics;
 - `UnleashServices` no longer initializes a `ToggleCollection` and `MetricsBucket`. Instead, it initializes an `YggdrasilEngine`;
 - `Variant` now inherits from `Yggdrasil.Variant`;
 - `Payload` now inherits from `Yggdrasil.Payload`;
 - `UnleashContext` now inherits from `Yggdrasil.Context`;
 - `ClientMetricsBackgroundTask` drops `ThreadSafeMetricsBucket` in favor of `YggdrasilEngine`;
 - `FetchFeatureTogglesTask` drops `ThreadSafeToggleCollection` and `IJsonSerializer` in favor of `YggdrasilEngine`;
 - Due to the simplification of some processes, like (de)serialization of state, some tests were simplified, adapted or removed. They can probably be simplified further on a later PR;

We also fixed some SDK behaviors that followed wrong assumptions.

For more details, feel free to ask or check the diff directly.


src/Unleash/DefaultUnleash.cs
```diff
@@ -105,130 +70,17 @@ public bool IsEnabled(string toggleName, UnleashContext context)
 
         public bool IsEnabled(string toggleName, UnleashContext context, bool defaultSetting)
         {
-            var enabled = CheckIsEnabled(toggleName, context, defaultSetting).Enabled;
```
sighphyre, line 108: Phew, seeing this ball of mud go away is very nice

src/Unleash/DefaultUnleash.cs
```diff
@@ -248,108 +100,17 @@ public Variant GetVariant(string toggleName, UnleashContext context)
...
-
-            services.MetricsBucket.RegisterCount(toggleName, variant.Name);
-        }
-
-        private static IStrategy[] SelectStrategies(IStrategy[] strategies, bool overrideDefaultStrategies)
-        {
-            if (overrideDefaultStrategies)
+            var variant = services.engine.GetVariant(toggleName, enhancedContext) ?? defaultValue;
+            
+            services.engine.CountVariant(toggleName, variant.Name);
```
sighphyre, line 107: There's some subtle stuff that may need love in the metrics for variants (I'm so sorry, the variant metrics have grown in weird ways).

- You need to count the feature yourself ala engine.CountToggle
- If the toggle doesn't exist, you should count the toggle as disabled but not the variant

(also thanks I realise there's a bug in the Ruby SDK Yggdrasil one, I count the disabled variant not the one specified by user - I think we don't support that in the UI so we probably have some work to do there as well)

src/Unleash/Internal/UnleashServices.cs
```diff
@@ -20,34 +18,43 @@ internal class UnleashServices : IDisposable
...
         internal CancellationToken CancellationToken { get; }
         internal IUnleashContextProvider ContextProvider { get; }
-        internal ThreadSafeToggleCollection ToggleCollection { get; }
         internal bool IsMetricsDisabled { get; }
-        internal ThreadSafeMetricsBucket MetricsBucket { get; }
         internal FetchFeatureTogglesTask FetchFeatureTogglesTask { get; }
-
-        public UnleashServices(UnleashSettings settings, EventCallbackConfig eventConfig, Dictionary<string, IStrategy> strategyMap)
+        internal YggdrasilEngine engine { get; }
+        private static readonly List<string> DefaultStrategyNames = new List<string> {
```
sighphyre, line 24:
```suggestion
        private static readonly IList<string> DefaultStrategyNames = new List<string> {
```

src/Unleash/Internal/UnleashServices.cs
```diff
@@ -20,34 +18,43 @@ internal class UnleashServices : IDisposable
...
             var cachedFilesResult = loader.EnsureExistsAndLoad();
 
-            ToggleCollection = new ThreadSafeToggleCollection
+            if (!string.IsNullOrEmpty(cachedFilesResult.InitialState))
             {
-                Instance = cachedFilesResult.InitialToggleCollection ?? new ToggleCollection()
-            };
-
-            MetricsBucket = new ThreadSafeMetricsBucket();
+                engine.TakeState(cachedFilesResult.InitialState);
```
sighphyre, line 56: What should we do if this fails? This can fail if the bootstrap are borked but I don't think this should bubble the exception all the way up - open to other thoughts here

sighphyre (commented):
This is looking sexy!

src/Unleash/Internal/UnleashServices.cs
nunogois: We discussed that the best idea we have for now is to wrap this in a try catch and log it. Addressed in https://github.com/Unleash/unleash-client-dotnet/pull/224/commits/4ff6c892d8de8824f21ae333fcc6585418a1fbe0

src/Unleash/DefaultUnleash.cs
nunogois: Addressed in https://github.com/Unleash/unleash-client-dotnet/pull/224/commits/63a3d68bdefa887b9fbafc6494649cb435018544

src/Unleash/Communication/UnleashApiClient.cs
```diff
@@ -211,29 +209,18 @@ public async Task<bool> RegisterClient(ClientRegistration registration, Cancella
...
+            jsonSerializer.Serialize(memoryStream, new
             {
-                jsonSerializer.Serialize(memoryStream, new ClientMetrics
-                {
-                    AppName = clientRequestHeaders.AppName,
-                    InstanceId = clientRequestHeaders.InstanceTag,
-                    Bucket = bucket
-                });
-            }
+                clientRequestHeaders.AppName,
```
sighphyre, line 220: Don't really mind if you wanna keep it but this used to be a named field and now it's not. Intentional?

src/Unleash/Internal/CachedFilesLoader.cs
```diff
@@ -72,7 +68,7 @@ public CachedFilesResult EnsureExistsAndLoad()
                 try
                 {
                     fileSystem.WriteAllText(toggleFile, string.Empty);
-                    result.InitialToggleCollection = null;
+                    result.InitialState = string.Empty;
```
sighphyre, line 71: For other reviewers - looks super suspect but this is really just to get a test to shush. All internal so I don't think it's a biggie

src/Unleash/Scheduling/ClientMetricsBackgroundTask.cs
```diff
@@ -1,37 +1,35 @@
...
-            this.metricsBucket = metricsBucket;
         }
 
         public async Task ExecuteAsync(CancellationToken cancellationToken)
         {
             if (settings.SendMetricsInterval == null)
                 return;
 
-            var result = await apiClient.SendMetrics(metricsBucket, cancellationToken).ConfigureAwait(false);
+            var result = await apiClient.SendMetrics(engine.GetMetrics(), cancellationToken).ConfigureAwait(false);
```
sighphyre, line 32: I think this is okay for now, we'll probably have to do another pass later in a different PR to deal with the fact that this completely drops metrics if the HTTP call borks - the GetMetrics call to Ygg is destructive in that it clears the metrics once they've been read.

Might be a "Metrics Extensions v3" problem

src/Unleash/Scheduling/FetchFeatureTogglesTask.cs
```diff
@@ -74,17 +71,17 @@ public async Task ExecuteAsync(CancellationToken cancellationToken)
             if (result.Etag == Etag)
                 return;
 
-            toggleCollection.Instance = result.ToggleCollection;
+            if (!string.IsNullOrEmpty(result.State))
+            {
+                engine.TakeState(result.State);
```
sighphyre, line 76: @daveleek This raises a new exception type, any idea where Exceptions that leave this method go? If they land in caller space then I'd like to wrap and reraise. If they get swallowed by async mechanics and become an error event then I don't care that much

src/Unleash/Unleash.csproj
```diff
@@ -16,31 +16,15 @@
   </PropertyGroup>
 
   <!-- Need to conditionally bring in references for the .NET Framework 4.* targets -->
- <ItemGroup Condition="'$(TargetFramework)' == 'net45'">
```
sighphyre, line 19: For other reviewers - we decided to drop older versions of .NET here. They're EOL, becoming burdensome to maintain and are blocking for the Yggdrasil wrapper

tests/Unleash.Tests/Internal/Dependent_Features_Tests.cs
```diff
@@ -124,12 +124,12 @@ public void Depends_On_One_Enabled_Parent_Counts_Metrics_Only_For_Child()
             var result = unleash.IsEnabled("child-1");
 
             // Assert
-            unleash.services.MetricsBucket.StopCollectingMetrics(out var bucket);
-            var childMetrics = bucket.Toggles.Single(t => t.Key == "child-1").Value;
-            childMetrics.No.Should().Be(0L);
-            childMetrics.Yes.Should().Be(1L);
+            var bucket = unleash.services.engine.GetMetrics();
+            var childMetrics = bucket?.Toggles.Single(t => t.Key == "child-1").Value;
+            childMetrics?.No.Should().Be(0L);
```
sighphyre, line 129: Oh that's very subtle, can we assert not null here? My eyes are telling me (perhaps wrong) that the asserts won't run if metrics is null

tests/Unleash.Tests/DefaultUnleashTests.cs
```diff
@@ -132,7 +132,13 @@ public static IUnleash CreateUnleash(string name, ToggleCollection state)
             var httpClient = new HttpClient(fakeHttpMessageHandler) { BaseAddress = new Uri("http://localhost") };
             var fakeScheduler = A.Fake<IUnleashScheduledTaskManager>();
             var fakeFileSystem = new MockFileSystem();
-            var toggleState = Newtonsoft.Json.JsonConvert.SerializeObject(state);
+            var toggleState = Newtonsoft.Json.JsonConvert.SerializeObject(state, new Newtonsoft.Json.JsonSerializerSettings
```
sighphyre, line 135: For other reviewers: this is a very clever little trick, compliments of @nunogois.

The existing state in the tests is now serialised and passed like that. This allows us to preserve the bulk of the tests with minimal changes and have confidence that we don't do too much surgery on the tests while we're also changing the code

The serialiser change here is to get .NET to send camelCased Json rather than PascalCased JSON, which Yggdrasil doesn'tt like

sighphyre (approved):
I'm happy. Few optional thoughts, mostly for others who may not have context coming into this PR.
I'm okay accepting this in it's current state, looks good! I do think we should wait for one more set of eyes though
Definitely worth another PR to get the full removal of the legacy business logic objects but let's merge this first. Great job!

src/Unleash/Communication/UnleashApiClient.cs
daveleek: So obviously not going to affect the outcome. But we should be consistent, we're using
`InstanceId = clientRequestHeaders.InstanceTag,` the row below so we should pick one and stick to it

src/Unleash/Communication/UnleashApiClient.cs
```diff
@@ -211,29 +209,18 @@ public async Task<bool> RegisterClient(ClientRegistration registration, Cancella
...
-        public async Task<bool> SendMetrics(ThreadSafeMetricsBucket metrics, CancellationToken cancellationToken)
+        public async Task<bool> SendMetrics(MetricsBucket metrics, CancellationToken cancellationToken)
         {
-            if (metricsRequestsToSkip > metricsRequestsSkipped)
-            {
-                metricsRequestsSkipped++;
-                return false;
-            }
-
-            metricsRequestsSkipped = 0;
```
daveleek, line 222: Why are we removing this?

src/Unleash/DefaultUnleash.cs
```diff
@@ -81,9 +49,6 @@ public DefaultUnleash(UnleashSettings settings, bool overrideDefaultStrategies,
             }
         }
 
-        /// <inheritdoc />
-        public ICollection<FeatureToggle> FeatureToggles => services.ToggleCollection.Instance.Features;
```
daveleek, line 85: Removing this is a breaking change, so should perhaps be pointed out in a major release. It's not in the documentation though, so nothing to update there

src/Unleash/IUnleash.cs
```diff
@@ -68,21 +61,6 @@ public interface IUnleash : IDisposable
...
-        /// <returns>A list of available variants.</returns>
-        IEnumerable<VariantDefinition> GetVariants(string toggleName);
-
-        /// <summary>
-        /// Get available feature variants.
-        /// </summary>
-        /// <param name="toggleName">The name of the toggle.</param>
-        /// /// <param name="context">The Unleash context to evaluate the toggle state against.</param>
-        /// <returns>A list of available variants.</returns>
-        IEnumerable<VariantDefinition> GetVariants(string toggleName, UnleashContext context);
```
daveleek, line 84: These aren't in the documentation either, but also warrants a major release

src/Unleash/Internal/Variant.cs
```diff
@@ -2,21 +2,18 @@
 
 namespace Unleash.Internal
 {
-    public class Variant
+    public class Variant: Yggdrasil.Variant
```
daveleek, line 5: What do we need to keep this one around for? I can't remember

src/Unleash/Unleash.csproj
daveleek: Great! We should be explicit about this and release a major version.
Seeing as we're publishing a [.NET Standard 2.0](https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0) theoretically we indirectly still support down to `4.6.2`

src/Unleash/Variants/VariantPayload.cs
```diff
@@ -1,31 +1,9 @@
-using System;
-
 namespace Unleash.Variants
 {
-    public class Payload
+    public class Payload: Yggdrasil.Payload
```
daveleek, line 3: Just wondering again why we have to keep this one around, I can't remember the exact reason

tests/Unleash.Tests/Internal/Dependent_Features_Tests.cs
daveleek: Yeah +1 on that

tests/Unleash.Tests/Utilities/ToggleBootstrapFileProvider_Tests.cs
```diff
@@ -49,8 +43,11 @@ public void Returns_File_Content_When_File_Exists()
             var result = toggleFileProvider.Read();
 
             // Assert
-            result.Features.Count().Should().Be(3);
-            result.Features.Single(f => f.Name == "featureY").Enabled.Should().Be(false);
+            result.Should().Be(fileContent);
+
+            var deserializedResult = Newtonsoft.Json.JsonConvert.DeserializeObject<ToggleCollection>(result);
+            deserializedResult?.Features.Count().Should().Be(3);
+            deserializedResult?.Features.Single(f => f.Name == "featureY").Enabled.Should().Be(false);
```
daveleek, line 50: Guess that goes for this one too, and the other test further down?

tests/Unleash.Tests/Utilities/ToggleBootstrapUrlProvider_Tests.cs
```diff
@@ -36,7 +31,8 @@ public void Gets_The_File_Content()
             var responseContent = bootstrapUrlProvider.Read();
 
             // Assert
-            responseContent.Features.Should().BeEmpty();
+            var deserializedResponseContent = JsonSerializer.Deserialize<ToggleCollection>(responseContent);
+            deserializedResponseContent?.Features.Should().BeEmpty();
```
daveleek, line 35: And these

src/Unleash/Scheduling/FetchFeatureTogglesTask.cs
daveleek: The primary use case is fine, gets called from the scheduler and outputs an error message in a log output. But this `ExecuteAsync` also gets called from `UnleashClientFactory.CreateClient` if `synchronousInitialization == true`, where errors are going straight back to caller space

daveleek (commented):
Awesome work, couple of +1's around some of the changes in tests, and a few comments on how we should communicate some of the contract changes

src/Unleash/Communication/UnleashApiClient.cs
nunogois: My editor complained about this one, due it to being redundant. I assume because in `AppName = clientRequestHeaders.AppName`, `AppName` can be deducted as the end property name, and as such it can be dropped.

That would not be the case with `InstanceId = clientRequestHeaders.InstanceTag` for example, since `InstanceId != InstanceTag`.

I'm OK with reverting this but I don't think it matters much 🤷

src/Unleash/Communication/UnleashApiClient.cs
nunogois: I have no idea what this is, but I saw it dropped in the new method you added in [#179](https://github.com/Unleash/unleash-client-dotnet/pull/179/files#diff-e189361ec95ef479c7161915691eb64f70a0ddaa41eac30285382906250fb0b7R223-R258) so I thought it was no longer relevant. Do we still care about this? Cc: @sighphyre

src/Unleash/DefaultUnleash.cs
nunogois: Yep, noted this in the PR description, and we're planning on a major release. Something that me and @sighphyre discussed is that we'll probably need to prepare a migration guide or similar, but out of scope for this PR.

src/Unleash/IUnleash.cs
nunogois: Yep, same as https://github.com/Unleash/unleash-client-dotnet/pull/224/files#r1661225276

src/Unleash/Internal/Variant.cs
nunogois: Well, `Variant` is a public class so I assume we still want it available. If we don't I'm more than happy to drop this entirely and just rely on `Yggdrasil.Variant` internally. WDYT?

src/Unleash/Scheduling/ClientMetricsBackgroundTask.cs
nunogois: Any chance this is related to https://github.com/Unleash/unleash-client-dotnet/pull/224/files#r1660524625?

src/Unleash/Unleash.csproj
nunogois: Correct, however like the doc you shared mentions:

> While NuGet considers .NET Framework 4.6.1 as supporting .NET Standard 1.5 through 2.0, there are several issues with consuming .NET Standard libraries that were built for those versions from .NET Framework 4.6.1 projects. For .NET Framework projects that need to use such libraries, we recommend that you upgrade the project to target .NET Framework 4.7.2 or higher.

src/Unleash/Variants/VariantPayload.cs
nunogois: Same as https://github.com/Unleash/unleash-client-dotnet/pull/224/files#r1661227880 perhaps?

tests/Unleash.Tests/Internal/Dependent_Features_Tests.cs
nunogois: Do you mean something like this? https://github.com/Unleash/unleash-client-dotnet/pull/224/commits/e7752b4284febb6d026b619f3571092a335004d1

Unfortunately my editor type checker doesn't care about the assertion and still complains that `bucket` could possibly be null, but that's probably okay.

image

src/Unleash/Scheduling/FetchFeatureTogglesTask.cs
nunogois: Is that intended? What should our plan be here?

tests/Unleash.Tests/Utilities/ToggleBootstrapFileProvider_Tests.cs
nunogois: What are these comments in reference to? 😄

nunogois: Oh you mean null assertion on the `deserializedResult`?

nunogois: Let me know if this is better: https://github.com/Unleash/unleash-client-dotnet/pull/224/commits/fe0f1cc65634e24d06fbaaaa1b14cd3e402aeaf8

tests/Unleash.Tests/Utilities/ToggleBootstrapUrlProvider_Tests.cs
nunogois: I assume this is the same as https://github.com/Unleash/unleash-client-dotnet/pull/224#discussion_r1660547726, addressed in the same commit: https://github.com/Unleash/unleash-client-dotnet/pull/224/commits/fe0f1cc65634e24d06fbaaaa1b14cd3e402aeaf8

src/Unleash/Communication/UnleashApiClient.cs
sighphyre: > My editor complained about this one

> I don't think it matters much 🤷

Yuh, happy for you to leave it alone if it's a tool complaining

src/Unleash/Scheduling/ClientMetricsBackgroundTask.cs
sighphyre: Kinda sorta. It's adjacent certainly. But I'll put details on that comment rather. This can still wait, the other we should talk about

src/Unleash/Communication/UnleashApiClient.cs
sighphyre: Ah, yeah, we should probably keep this for now. This is the HTTP backoff stuff we added to stop hammering Unleash when it's not ready to handle a ton of traffic I think this is dropped in #179 because that's quite old (better to say 179 didn't know that it needed it yet because it didn't exist).

This probably needs to change at some point in the near future but this might _also_ be a metrics v3 problem

src/Unleash/Scheduling/FetchFeatureTogglesTask.cs
sighphyre: > The primary use case is fine, gets called from the scheduler and outputs an error message in a log output. But this `ExecuteAsync` also gets called from `UnleashClientFactory.CreateClient` if `synchronousInitialization == true`, where errors are going straight back to caller space

If I'm reading that correctly then yes, lands in user space

@nunogois sounds like 

```try{ 
  engine.TakeState(result.State);
} catch(YggException ex){
  throw new UnleashException(ex)
}```

or something very close is the correct thing here. Effectively what I don't want to happen here is code in the wild that's only catching `UnleashException` to have an YgggrasilException blow past it and unwind someone's stack

File: src/Unleash/Unleash.csproj (hunk 9 above)
Comment (sighphyre) [reply src/Unleash/Unleash.csproj:19]: @daveleek If I read @nunogois and the linked doc correctly (and I only _think_ I do) then `4.6.2` is out. I'm open to being told we need `4.6.2` but burdensome is what I want to call that and I'm very lazy

File: tests/Unleash.Tests/Internal/Dependent_Features_Tests.cs (hunk 10 above)
Comment (sighphyre) [reply tests/Unleash.Tests/Internal/Dependent_Features_Tests.cs]:
```
bucket.Should().NotBeNull();
```
Yuh. Happy

File: src/Unleash/Internal/Variant.cs (hunk 15 above)
Comment (sighphyre) [reply src/Unleash/Internal/Variant.cs:5]: >  What do we need to keep this one around for? I can't remember

In theory, it stops people having to run around like crazy people updating their namespaces. Just makes this upgrade a slightly less bitter pill to swallow

File: src/Unleash/Scheduling/FetchFeatureTogglesTask.cs (hunk 8 above)
Comment (nunogois) [reply src/Unleash/Scheduling/FetchFeatureTogglesTask.cs]: Thanks! How about something like this? https://github.com/Unleash/unleash-client-dotnet/pull/224/commits/29650ef55ba60863197fb2f587350be5180acb44 - Should be consistent with how other exceptions are handled in this file.

File: src/Unleash/Communication/UnleashApiClient.cs (hunk 12 above)
Comment (nunogois) [reply src/Unleash/Communication/UnleashApiClient.cs]: Thanks for clarifying. https://github.com/Unleash/unleash-client-dotnet/pull/224/commits/61ebe1716a59188da61aca70c763562252fac2e9 brings this back for now.

File: src/Unleash/Communication/UnleashApiClient.cs (hunk 5 above)
Comment (daveleek) [reply src/Unleash/Communication/UnleashApiClient.cs:228]: Fair point, I'm not going to be too fussy about it either then :)

File: src/Unleash/Unleash.csproj (hunk 9 above)
Comment (daveleek) [reply src/Unleash/Unleash.csproj:19]: No I'm fine with them being dropped by us which is what I'm saying in a roundabout way. With Standard 2.0 we're supporting a few runtimes (through using Standard 2.0) even if we drop the explicit support for them, in the cases where that is absolutely necessary. But yes, people on 4.6.1 might experience issues trying to add our library to their projects

File: tests/Unleash.Tests/Internal/Dependent_Features_Tests.cs (hunk 10 above)
Comment (daveleek) [reply tests/Unleash.Tests/Internal/Dependent_Features_Tests.cs]: Yes, happy!

File: src/Unleash/Internal/Variant.cs (hunk 15 above)
Comment (daveleek) [reply src/Unleash/Internal/Variant.cs:5]: Ok if you're happy keeping it around then sure. It could be part of the migration article mentioned above though, since they're already having to do a bunch of work

File: src/Unleash/Variants/VariantPayload.cs (hunk 16 above)
Comment (daveleek) [reply src/Unleash/Variants/VariantPayload.cs:3]: yup

File: tests/Unleash.Tests/Utilities/ToggleBootstrapFileProvider_Tests.cs (hunk 17 above)
Comment (daveleek) [reply tests/Unleash.Tests/Utilities/ToggleBootstrapFileProvider_Tests.cs]: Apologies for lack of context there, but yes! and thanks!

File: src/Unleash/Scheduling/FetchFeatureTogglesTask.cs (hunk 8 above)
Comment (daveleek) [reply src/Unleash/Scheduling/FetchFeatureTogglesTask.cs]: Awesome! Thanks @nunogois 👌

**Review by daveleek — APPROVED** [review]
Awesome work @nunogois!

**Comment by nunogois** [comment]
Rebased this and cleaned it up a bit.
@sighphyre let me know when you think it's a good time to merge this.

File: src/Unleash/DefaultUnleash.cs [hunk 19]
```diff
@@ -2,61 +2,32 @@ namespace Unleash
...
         internal readonly UnleashServices services;
 
         private readonly WarnOnce warnOnce;
 
-        ///// <summary>
-        ///// Initializes a new instance of Unleash client with a set of default strategies.
-        ///// </summary>
-        ///// <param name="config">Unleash settings</param>
-        ///// <param name="strategies">Additional custom strategies.</param>
-        public DefaultUnleash(UnleashSettings settings, params IStrategy[] strategies)
```
daveleek, line 49: Replacing this with an Ygg strategy list is a breaking change

src/Unleash/Internal/IToggleBootstrapProvider.cs
```diff
@@ -1,11 +1,7 @@
...
-using System.Collections.Generic;
-using System.Text;
-
-namespace Unleash.Internal
+﻿namespace Unleash.Internal
 {
     public interface IToggleBootstrapProvider
     {
-        ToggleCollection Read();
+        string Read();
```
daveleek, line 5: Breaking change

daveleek: Let's change the name of this one completely. Also now has a deprecation notice in main

rcollette:
What architectures are going to be supported in docker containers?  I've been getting burned by rust/go .NET integrations lately where the producing teams do not have access to ARM64 platforms on which to build and code that runs on my local machine, fails to build in a docker on ARM64 Macs.

tests/Unleash.Tests/DefaultUnleashTests.cs
rcollette: If making a major (semver) change for this release, why not go to System.Text.Json, which would then allow for further non-breaking work to adopt JSON source generators.

sighphyre:
> What architectures are going to be supported in docker containers? I've been getting burned by rust/go .NET integrations lately where the producing teams do not have access to ARM64 platforms on which to build and code that runs on my local machine, fails to build in a docker on ARM64 Macs.

Out the box, likely x86 Windows/MacOS/Linux + ARM MacOS. Right now there's no Github runners for ARM Windows/Linux, not a blocker but we likely won't go through the dance of setting up a custom runner unless someone asks

tests/Unleash.Tests/DefaultUnleashTests.cs
sighphyre: Yep. we chatted, it's a good idea, we're gonna go all in on System.Text.Json. Not in this PR though, this one's already too big

PR #226, opened 2024-07-08
feat: automatically generated instanceId

nunogois:
https://linear.app/unleash/issue/2-2393/make-instance-id-autogenned-c

This drops the `InstanceTag` public property in `UnleashSettings` in favor of an internal automatically generated `InstanceId` property. Any reference of `InstanceTag` is now `InstanceId`.

File names no longer include `InstanceTag` and instead we're relying on [gracefully handling file locks](https://github.com/Unleash/unleash-client-dotnet/blob/main/tests/Unleash.Tests/IOTests.cs).


README.md
```diff
@@ -425,24 +424,7 @@ By default unleash-client fetches the feature toggles from unleash-server every
...
-    // Set an instance tag for consistent backup file naming
-    InstanceTag = "CustomInstanceTag",
-    UnleashContextProvider = new AspNetContextProvider(),
-    CustomHttpHeaders = new Dictionary<string, string>()
-    {
-        {"Authorization", "API token" }
-    }
-};
-```
+The backup file name will follow this pattern: `{fileNameWithoutExtension}-{AppName}-{SdkVersion}.{extension}`.
```
sighphyre, line 427: Probably worth mentioning that we took this out and why

sighphyre:
How does this play with this:

> https://github.com/Unleash/unleash-client-dotnet/blob/93f4d88a1058581960752d754954dee665a16435/src/Unleash/Scheduling/FetchFeatureTogglesTask.cs#L92

If I'm reading that right this is still spitting out an error onto the error events. I think it shouldn't do that anymore WDYT?

sighphyre (approved):
I think we should change the error event but I don't think it's dangerous if you disagree so I'm gonna put an LGTM on it

nunogois:
> How does this play with this:
> 
> > https://github.com/Unleash/unleash-client-dotnet/blob/93f4d88a1058581960752d754954dee665a16435/src/Unleash/Scheduling/FetchFeatureTogglesTask.cs#L92
> 
> If I'm reading that right this is still spitting out an error onto the error events. I think it shouldn't do that anymore WDYT?

Yep, addressed in https://github.com/Unleash/unleash-client-dotnet/pull/226/commits/78cf8c291e8d99d03bbaf9b80c69a30f74f714a8

PR #227, opened 2024-07-08
chore: make metrics and regsitration tests actually useful

sighphyre:
Makes the metrics and registration tests actually do something. They're previously been disabled they hit a heroku instance with an invalid key and... well... didn't do anything useful. This wires them up to a mock HTTP implementation and validates that the expected information gets send to Unleash on registration and metrics heartbeat

I've also chosen to discard one feature toggle test that appears to be testing system timers in C#. Open to bringing it back if there's value I missed there


tests/Unleash.Tests/Communication/UnleashApiClient_FetchToggles_Tests.cs
```diff
@@ -1,53 +0,0 @@
...
-            // With etag from previous response
-            var result2 = await api.FetchToggles(result1.Etag, CancellationToken.None);
-            result2.HasChanged.Should().BeFalse();
-            result2.ToggleCollection.Should().BeNull();
-            result2.Etag.Should().Be(result1.Etag);
-        }
-
-        [Test]
-        [Ignore("Requires a valid accesstoken")]
-        public async Task Timer_Test()
```
sighphyre, line 37: Why?

tests/Unleash.Tests/Communication/MockHttpClient.cs
```diff
@@ -0,0 +1,37 @@
+﻿using Unleash.Communication;
+using Unleash.Internal;
+using Unleash.Serialization;
+using RichardSzalay.MockHttp;
+
+namespace Unleash.Tests.Communication
+{
+    internal static class MockHttpClient
+    {
+        internal static Tuple<MockHttpMessageHandler, UnleashApiClient> MakeMockClient(string url)
```
sighphyre, line 10: It's 2024, we can just call functions rather than relying on an inheritance chain for test setup

nunogois (approved):
LGTM

PR #228, opened 2024-07-09
chore: put an autoformatter in

sighphyre:
This has been sorely lacking for a while. Getting a bit tired of playing formatter ping pong so now it's enforced by a tool in the build

This PR does 3 things:

- Applies `dotnet format` to the whole codebase
- Adds instructions to the README for installing the tool yourself if you don't already have it
- Runs verify no changes in the build to check that the formatter was correctly applied to incoming PRs

PR #229, opened 2024-07-09
chore: reduce noisy warning and error output during test runs

sighphyre:
Assumes #228 has been applied, otherwise the diff will be very noisy.

This reduces the noisy warning overhead in the test runs. These is caused by a few errors and warnings which add little to no value at test time:

- The test system tends to reuse Unleash instances in the way that we specifically warn users not to. That spits out the instance warning on most test runs. For this I've added a way to silence the singleton warnings in the settings (other SDKs have this so I don't feel bad). This defaults to off so it shouldn't change behavior
- One of the tests just has super noisy output. I've chosen to remove that, no one was ever reading it


nunogois (approved):
Oh yes, huge +1 for more readable test outputs.

PR #230, opened 2024-07-09
chore: add ignore for vs code backup build file

sighphyre:
It's driving me mad


nunogois (approved):
Yep. I guess this means we're committing the new task files that are generated.

sighphyre:
> Yep. I guess this means we're committing the new task files that are generated.

I think they're already committed! This is just me trying to change as little as possible but also not manage my staging area manually every 5 minutes because vscode and dotnet have chosen violence

PR #231, opened 2024-07-09
feat: extend metrics with metadata

sighphyre:
Needs #229, otherwise it'll be noisy

This adds metadata to the metrics and registration calls. This has a few sharp edges to it:

- Insofar as I can tell, there's no reliable way to get the version of the CLR or framework when running in C#. Everything has some details where it won't work on something. I've chosen to go with some degenerate string splitting + a compiler directive for older versions, which should give us enough information to know what's happening in the wild and always return "something"
- The design of the way async tasks work in this SDK needs to be rewritten to reduce the coupling between async stuff and business logic stuff. Right now there's no clean way to separate these two, which means there's no good way to extend the metrics/registration in a way that isn't imminently breakable from the consumer perspective or testable in any rational way. Because of this, I've chosen to make these properties lazy getters on the Metrics/Registration objects themselves because it's hard to accidentally break and it gives us some testing

The task system needs some love but I don't think it's responsible to do that before Yggdrasil gets merged

### DIFF
--- a/src/Unleash/Internal/UnleashServices.cs
+++ b/src/Unleash/Internal/UnleashServices.cs
@@ -16,7 +16,7 @@ namespace Unleash
         private readonly CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
         private readonly IUnleashScheduledTaskManager scheduledTaskManager;
 
-        const string supportedSpecVersion = "4.5.1";
+        public const string supportedSpecVersion = "4.5.1";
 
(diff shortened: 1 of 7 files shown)


nunogois (approved):
Shame there's no better way of getting this information, but this looks good 👍

sighphyre:
> Shame there's no better way of getting this information, but this looks good 👍

+1! I think we can make it more complex to get richer information but it doesn't feel worth it right now

PR #232, opened 2024-07-11
chore: cleanup post chore-yggdrasil

nunogois:
Follow up to: https://github.com/Unleash/unleash-client-dotnet/pull/224

Major cleanup, removing things we no longer need.

It's not extremely exhaustive, as there are more things that can probably be cleaned up at a later stage.

Spec tests are most likely failing because we'll need: https://github.com/Unleash/yggdrasil/pull/134


src/Unleash/Metrics/BaseMetrics.cs
```diff
@@ -0,0 +1,37 @@
+namespace Unleash.Metrics
```
sighphyre, line 1: Oh cheers, nice catch

src/Unleash/DefaultUnleash.cs
```diff
@@ -118,7 +119,7 @@ public Variant GetVariant(string toggleName, UnleashContext context, Variant def
                 EmitImpressionEvent("getVariant", enhancedContext, variant.IsEnabled, toggleName, variant.Name);
             }
 
-            return Variant.UpgradeVariant(variant);
+            return variant;
```
sighphyre, line 122: I don't think I'm sold on this one. Leaking a child dependency through a public API seems... well... very leaky. I don't think the cost of keeping just the Variant class around is particularly high. This PR is already very ambitious in the scope of things it removes... can we just keep it for now?

sighphyre (commented):
I'm mostly sold. I think the bootstrapping tests are mostly useless now but I'd rather keep them until we replace them with something better so they can just sit there.

I would like it if Ygg doesn't leak through the public API though. I can deal with it for strategies but variants feels like a step too far personally. I don't think the cost is too high to just keep variants. This is still a -6K line PR (whooo!)

src/Unleash/DefaultUnleash.cs
nunogois: Yep, after we've discussed this, I brought it back in https://github.com/Unleash/unleash-client-dotnet/pull/232/commits/c11ce8114b79642fc522f6ee6f4c2bedb151a877

PR #235, opened 2024-08-12
feat: always reset stream position after call to serializer

### LINKED ISSUE #222: Avoid subtle errors and improve DX when setting a custom serializer
**Is your feature request related to a problem? Please describe.**
Recently we switched to using System.Text.Json and we fell in the same error as #197 and it went unoticed until production because no one knew that we had to set `stream.Position = 0` in the serializer.
We might have it in the unleash doc, but I think even that won't fix the real root cause. 

**Describe the solution you'd like**
I think it would be better to set the stream.Position inside the method that is calling the serializer (when needed), this way we won't have the developer to set it at the end of the Serialize method of the serializer. 

**Proposed implementation**
add `stream.Position` = 0 here : 
https://github.com/Unleash/unleash-client-dotnet/blob/93f4d88a1058581960752d754954dee665a16435/src/Unleash/Communication/UnleashApiClient.cs#L237

Again, if this is a good way to do it and I'm not missing context, I will be very happy to submit a PR. 

Thanks 😊

daveleek:
Resets stream position to 0 every time a call to serializer has been made.

Fixes #222 

Tested this locally, should work fine with existing code bases already calling stream.Position = 0 on their serializer implementations

### DIFF
--- a/src/Unleash/Communication/UnleashApiClient.cs
+++ b/src/Unleash/Communication/UnleashApiClient.cs
@@ -187,6 +187,7 @@ namespace Unleash.Communication
 
             var memoryStream = new MemoryStream();
             jsonSerializer.Serialize(memoryStream, registration);
+            memoryStream.Position = 0;
 
             const int bufferSize = 1024 * 4;
 
@@ -233,6 +234,7 @@ namespace Unleash.Communication
                     InstanceId = clientRequestHeaders.InstanceTag,
                     Bucket = bucket
                 });
+                memoryStream.Position = 0;
             }
 
             const int bufferSize = 1024 * 4;
--- a/README.md
+++ b/README.md
@@ -533,7 +533,6 @@ public class NewtonsoftJson7Serializer : IJsonSerializer
             Serializer.Serialize(jsonWriter, instance);
 
             jsonWriter.Flush();
-			stream.Position = 0;
         }
     }
 


src/Unleash/Communication/UnleashApiClient.cs
```diff
@@ -233,6 +234,7 @@ public async Task<bool> SendMetrics(ThreadSafeMetricsBucket metrics, Cancellatio
                     InstanceId = clientRequestHeaders.InstanceTag,
                     Bucket = bucket
                 });
+                memoryStream.Position = 0;
```
gastonfournier, line 237: Should we also update the docs?

Also, what we discussed is why should we allow people to override the json serialization mechanism? Maybe that feature should be deprecated

gastonfournier (approved):
LGTM!

src/Unleash/Communication/UnleashApiClient.cs
daveleek: > Should we also update the docs?

Yes, I'll amend this PR

> Also, what we discussed is why should we allow people to override the json serialization mechanism? Maybe that feature should be deprecated

Absolutely, but feels like a bigger question/release since we're changing the public API. I figured this would be a small fix in the meantime while we wrap up what we're working on for the next major release (Ygg++)

Do you agree?

gastonfournier: Agree!

PR #236, opened 2024-09-02
chore: remove unsupported .NET framework versions

daveleek:
In preparation for [Yggdrasil PR](https://github.com/Unleash/unleash-client-dotnet/pull/224), remove all .NET framework versions < 4.7.2. We're keeping .NET Standard 2.0


sighphyre (approved):
Sounds like we can go all the way to just supporting netstandard2.0 and maybe net6.0 but looks good. I'm going to put an approve so you don't get blocked on this

PR #237, opened 2024-09-02
chore: prepare for Yggdrasil with deprecations of changing API

daveleek:
Deprecations in preparation of [Yggdrasil PR](https://github.com/Unleash/unleash-client-dotnet/pull/224) that removes some public APIs

### DIFF
--- a/src/Unleash/Utilities/ToggleBootstrapUrlProvider.cs
+++ b/src/Unleash/Utilities/ToggleBootstrapUrlProvider.cs
@@ -30,8 +30,7 @@ namespace Unleash.Utilities
             this.throwOnFail = throwOnFail;
             this.customHeaders = customHeaders;
         }
-
-        [Obsolete("Will return json string in the next major version", false)]
+        [Obsolete("Will be replaced in the next major version", false)]
         public ToggleCollection Read()
         {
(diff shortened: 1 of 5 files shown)


src/Unleash/Utilities/ToggleBootstrapUrlProvider.cs
```diff
@@ -31,6 +31,7 @@ public ToggleBootstrapUrlProvider(string path, HttpClient client, UnleashSetting
             this.customHeaders = customHeaders;
         }
 
+        [Obsolete("Will return json string in the next major version", false)]
```
sighphyre, line 34: Think we should do a different method rather than changing the return type

sighphyre (approved):
Yeah LGTM

rcollette:
What is the replacement for `.FeatureToggles`?

sighphyre:
> What is the replacement for `.FeatureToggles`?

There likely won't be a direct replacement. To be completely honest there's still some debate to be had here. The two options (not necessarily competing) are to either return a raw JSON string, which is the unparsed response directly from the server, or a list of names of each of the feature toggles known.

IMO exposing the feature toggles directly was never a very good idea, outside of exposing their names.

I'm open to suggestions if you're using this in some way we hadn't considered yet though

PR #238, opened 2024-09-03
chore: use method signature overloads instead of dynamic in scheduler

daveleek:
Removes the dynamic parameters and their required type checks from the scheduler extensions and replaces with overloads for the two types we do expect as parameters: int and TimeSpan


sighphyre (approved):
Uh... yes please!

PR #239, opened 2024-09-04
chore: trim exported platform dlls to standard 2.0 only

daveleek:
In preparation for Yggdrasil trim down exported platform dlls to only Standard 2.0.
Removes conditional imports for .NET Framework versions and makes permanent import for Microsoft.CSharp package


sighphyre (approved):
Nice!

PR #241, opened 2024-09-19
fix: increase the possible random numbers used for stickiness id

thomasheartman:
Same fix as done in some other sdks, such as the node one at
https://github.com/Unleash/unleash-client-node/pull/417

Fixes an issue where 1% rollout would not yield any results with
random rollout for certain group ids


sighphyre (approved):
LGTM

sighphyre (commented):
Wait no. Wait this looks weird. Random.Next() returns a 32 bit integer. I don't think multiplying it by anything is going to help here. I think that's just going to make it overflow and do weird things

Kinda think this one maybe needs a test

thomasheartman:
Yeh, replaced it with a bounded random between 1 (inclusive) and 10,001 (exclusive) instead of doing the same math as before.

thomasheartman:
@sighphyre Added a test that appears to be working fine. You good with merging it?

PR #242, opened 2024-09-24
chore: change some header levels in README.md

thomasheartman:
The formatting header was its own top-level header with "release
process" nested under it. I'm assuming it should've been one level
deeper (the same as release process).

Also, "other information" isn't really development stuff, so let's
raise it.


daveleek (approved):
Awesome! :shipit:

PR #246, opened 2024-10-23
Fix currentTime constraint when CurrentTime is not explicitly set

### LINKED ISSUE #245: currentTime strategy constraint doesn't work
**Describe the bug**
If I create a strategy with a `DATE_AFTER` constraint on `currentTime`, with a date in the past, the feature flag is always considered disabled.
The same feature flag works fine on the frontend with [unleash-proxy-client-js](https://github.com/unleash/unleash-proxy-client-js).

**To Reproduce**
1. Create a feature flag with a gradual rollout strategy with 100% rollout
2. Add a constraint on `currentTime` with `DATE_AFTER` and a date in the past
3. In the app, call `IUnleash.IsEnabled("MyFeatureFlag")`
4. It always returns false, whether the current date is before or after the specified date
5. Remove the `currentTime` constraint
6. Now `IUnleash.IsEnabled("MyFeatureFlag")` returns true

**Expected behavior**
`IUnleash.IsEnabled("MyFeatureFlag")` should return true if the current date is after the specified date

**Screenshots**
image

**Desktop (please complete the following information):**
 - OS: Ubuntu (deployed application), Windows (for local development)
 - Browser: irrelevant (server application)
 - Version: 4.1.13

tlevesque-ueat:
# Description

Change `UnleashContext.GetByName` to return `CurrentTime` when `contextName` is `"currentTime"` (or `DateTimeOffset.UtcNow` when `UnleashContext.CurrentTime` is not set).

Also, I suspect the `else` branch of `if (!string.IsNullOrWhiteSpace(contextValue))` in `DateConstraintOperator.Evaluate` is wrong: I don't see why we would use `context.CurrentTime` if there is no value for the `ContextName` specified in the constraint. It should only be done if the `ContextName` is `"currentTime"` (but it would be redundant anyway after my change). I think that whole `else` branch should be removed. I didn't make that change yet, wanted to check with you guys first.

Fixes #245

## Type of change

# How Has This Been Tested?

Unit tests

**Test Configuration**:
* Firmware version: not relevant
* Hardware: not relevant
* Toolchain: not relevant
* SDK: .NET SDK 8.0.403

# Checklist:

### DIFF
--- a/src/Unleash/UnleashContext.cs
+++ b/src/Unleash/UnleashContext.cs
@@ -46,6 +46,8 @@ namespace Unleash
                     return SessionId;
                 case "remoteAddress":
                     return RemoteAddress;
+                case "currentTime":
+                    return (CurrentTime ?? DateTimeOffset.UtcNow).ToString("O");
                 default:
                     string result;
                     Properties.TryGetValue(contextName, out result);
--- a/tests/Unleash.Tests/Strategy/Constraints/DateConstraintOperator_Tests.cs
+++ b/tests/Unleash.Tests/Strategy/Constraints/DateConstraintOperator_Tests.cs
@@ -147,5 +147,35 @@ namespace Unleash.Tests.Strategy.Constraints
             // Assert
             result.Should().BeTrue();
         }
+
+        [Test]
+        public void DATE_AFTER_CurrentTime()
+        {
+            // Arrange
+            var target = new DateConstraintOperator();
+            var constraint = new Constraint("currentTime", Operator.DATE_AFTER, false, false, DateTime.UtcNow.AddDays(-30).ToString("O"));
+            var context = new UnleashContext.Builder().Now().Build();
+
+            // Act
+            var result = target.Evaluate(constraint, context);
+
+            // Assert
+            result.Should().BeTrue();
+        }
+
+        [Test]
+        public void DATE_AFTER_CurrentTime_Not_Set_Uses_UtcNow()
+        {
+            // Arrange
+            var target = new DateConstraintOperator();
+            var constraint = new Constraint("currentTime", Operator.DATE_AFTER, false, false, DateTime.UtcNow.AddDays(-30).ToString("O"));
+            var context = new UnleashContext();
+
+            // Act
+            var result = target.Evaluate(constraint, context);
+
+            // Assert
+            result.Should().BeTrue();
+        }
     }
 }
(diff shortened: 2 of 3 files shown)


sighphyre (approved):
Thanks for catching this! LGTM!

tlevesque-ueat:
Thanks for the merge!

What about this? It looks like a bug tbh...



> Also, I suspect the `else` branch of `if (!string.IsNullOrWhiteSpace(contextValue))` in `DateConstraintOperator.Evaluate` is wrong: I don't see why we would use `context.CurrentTime` if there is no value for the `ContextName` specified in the constraint. It should only be done if the `ContextName` is `"currentTime"` (but it would be redundant anyway after my change). I think that whole `else` branch should be removed. I didn't make that change yet, wanted to check with you guys first.

daveleek:
@tlevesque-ueat 
> What about this? It looks like a bug tbh...

There's a similar fallback in the reference SDKs used for this implementation (Node, Java). You might be right in that it should only do this for `"currentTime"` though

sighphyre:
> Thanks for the merge!
> 
> What about this? It looks like a bug tbh...
> 
> > Also, I suspect the `else` branch of `if (!string.IsNullOrWhiteSpace(contextValue))` in `DateConstraintOperator.Evaluate` is wrong: I don't see why we would use `context.CurrentTime` if there is no value for the `ContextName` specified in the constraint. It should only be done if the `ContextName` is `"currentTime"` (but it would be redundant anyway after my change). I think that whole `else` branch should be removed. I didn't make that change yet, wanted to check with you guys first.

+1, that does look like a bug to my eyes as well but no longer relevant after your change. If you're up for doing a PR for that, I can have a release out with both changes today?

sighphyre:
> @tlevesque-ueat
> 
> > What about this? It looks like a bug tbh...
> 
> There's a similar fallback in the reference SDKs used for this implementation (Node, Java). You might be right in that it should only do this for `"currentTime"` though

I can see what the code was intended to do, based on Node but I don't think it does that

sighphyre:
@tlevesque-ueat Actually, I'm being silly, I can do that quickly and do a release. Should be out in a few mins

tlevesque-ueat:
> +1, that does look like a bug to my eyes as well but no longer relevant after your change.

How so? If you create a constraint like `myCustomField DATE_AFTER <some date>`, and `myCustomField` is not set, it will still fall back to using `context.CurrentTime` instead, which is arguably wrong.

tlevesque-ueat:
> > +1, that does look like a bug to my eyes as well but no longer relevant after your change.
> 
> How so? If you create a constraint like `myCustomField DATE_AFTER <some date>`, and `myCustomField` is not set, it will still fall back to using `context.CurrentTime` instead, which is arguably wrong.

Never mind, I just noticed you already removed the `else` in #247 anyway! 👍

PR #247, opened 2024-10-25
chore: remove unnecessary else branch in date constraint

sighphyre:
This just removes the `else` statement code here and makes the `if` the only code path.

The `else` is now redundant, based on this PR #246 which forces the context to always have a current time


gastonfournier (approved):
Looks good!

Issue #221, opened 2024-05-15
Built-in JSON serialization without reflection


gastonfournier:
**Is your feature request related to a problem? Please describe.**
Depending on a [Json serialization](https://github.com/Unleash/unleash-client-dotnet?tab=readme-ov-file#json-serialization) library that requires a custom implementation, can lead to errors. In particular, when defining a custom serialization with system.text if we forget to do `stream.Position = 0;` the payload will not be serialized properly, leading to errors downstream.

**Describe the solution you'd like**
Serialization should be provided by the SDK, it's not something users should override, as there's only 1 correct way of serializing the payload. The downside is that the actual users, that are using a custom serializer might cause a crash at runtime.

**Describe alternatives you've considered**
None, but open to suggestions

**Additional context**
As a sample test, I wrote a test for System.text as serializer. It's currently not working because the test compares strings to strings and the only difference is the sort order of fields. So it's the test that needs to change. On top of that, the test is not covering serialization of metrics, so it's incomplete. But it's a good enough to prove the point that serialization is dangerous if left in the hands of users and it should be handled by the SDK without giving users the ability to override it:

https://github.com/Unleash/unleash-client-dotnet/compare/system-text-serialization-test?expand=1

gastonfournier:
We're making this issue visible after some issues caused by custom serializers, but it's not gonna be prioritized soon. We'd like to collect some feedback from the community (you), and why not some contribution.

PhbsSmn:
Why not opting for System.Text.Json rather than still using the Newtonsoft implementation?
Did not glance too deeply in the code base but for the things I needed it wasn't such a hard thing, and by doing so you could tap into the power of source generated serialization as well for the newer .NET versions.

Not sure how backwards compatible this is though since it's been a long time now that I used .NET framework.

sighphyre:
> Why not opting for System.Text.Json rather than still using the Newtonsoft implementation? Did not glance too deeply in the code base but for the things I needed it wasn't such a hard thing, and by doing so you could tap into the power of source generated serialization as well for the newer .NET versions.
> 
> Not sure how backwards compatible this is though since it's been a long time now that I used .NET framework.

Yeah I think this makes sense

sighphyre:
Handled by #249

sighphyre:
This should be out in the next major release

PR #248, opened 2024-10-25
chore: add deprecation warning to settable json serializer

sighphyre:
Does what it says on the tin, first steps towards closing #221

### DIFF
--- a/src/Unleash/UnleashSettings.cs
+++ b/src/Unleash/UnleashSettings.cs
@@ -103,6 +103,7 @@ namespace Unleash
         ///
         /// Default: A serializer based on Newtonsoft will be used, given that these assemblies are loaded into the appdomain already.
         /// </summary>
+        [Obsolete("No longer supported in recent versions of Unleash, will be removed in the next major version", false)]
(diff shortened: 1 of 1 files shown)


src/Unleash/UnleashSettings.cs
```diff
@@ -103,6 +103,7 @@ public class UnleashSettings
         ///
         /// Default: A serializer based on Newtonsoft will be used, given that these assemblies are loaded into the appdomain already.
         /// </summary>
+        [Obsolete("No longer supported in recent versions of Unleash, will be removed in the next major version", false)]
```
nunogois, line 106: It's not really "recent versions of Unleash" though, is it? How about something like this instead?

```suggestion
        [Obsolete("This setting is no longer supported in recent versions of the Unleash .NET SDK and will be removed in the next major release.", false)]
```

nunogois (approved):
Yes please 👍

src/Unleash/UnleashSettings.cs
sighphyre: It's just the same deprecation notice that got splashed on a bunch of other things in preparation for the next version. I'm going to keep it consistent

PR #249, opened 2024-10-28
feat: move to System.Text.Json as serialiser

sighphyre:
This removes the deprecated option to set your own JSON serializer. This should be very low impact on consumers, since this just requires them to remove a custom serializer if one had previously be set. The SDK now uses `System.Text.Json` which is built into most modern versions of .NET and is added as a dependency for legacy versions that don't support it natively.

Almost all of this PR is noise removing references, I'll mark relevant sections with some comments

## Discussion

There's some, frankly, weird stuff around streams in the SDK at the moment. I'm 99% sure that's an artifact of the old, slightly awkward interfaces. I've chosen to throw that out, write directly and let the HTTP client deal with the buffering. I'm pretty sure that's not a big loss since the old implementation was flushing to the streams multiple times and the new implementation can use async to not have to worry too much about what's being blocked


src/Unleash/Communication/UnleashApiClient.cs
```diff
@@ -183,15 +187,9 @@ public async Task<bool> RegisterClient(ClientRegistration registration, Cancella
...
-            var memoryStream = new MemoryStream();
-            jsonSerializer.Serialize(memoryStream, registration);
-            memoryStream.Position = 0;
-
-            const int bufferSize = 1024 * 4;
-
             using (var request = new HttpRequestMessage(HttpMethod.Post, requestUri))
             {
-                request.Content = new StreamContent(memoryStream, bufferSize);
+                request.Content = new StringContent(JsonSerializer.Serialize(registration, options), Encoding.UTF8, "application/json");
```
sighphyre, line 192: Swap out new streaming. This should have a default buffer of 1024 * 4 bytes anyway

src/Unleash/Unleash.csproj
```diff
@@ -71,6 +71,7 @@
     </PackageReference>
     <PackageReference Include="murmurhash" Version="1.0.3" />
     <PackageReference Include="NuGet.Versioning" Version="6.1.0" />
+    <PackageReference Include="System.Text.Json" Version="8.0.5" />
```
sighphyre, line 74: Force a reference to System.Text.Json for older versions that don't have it built in

daveleek (approved):
This is great! You have my blessing, feels good to get some of this complexity removed. A little paranoid testing as we've already discussed, for instance before and after this change with backups created by one and then loaded by the other might not be a bad idea.

sighphyre:
@daveleek I'm less worried about loading features from file, that's effectively a cache miss on major upgrade which feels okay. Bit worried about Edge so agree we need to validate this more before release

We can play tomorrow and try break it!

PR #251, opened 2024-10-30
chore: remove legacy domain entities and associated tests

sighphyre:
Trims out the legacy artifacts, tries to make minimal changes to tests to preserve sanity. None of this is needed post Yggdrasil and needs to be removed prior to next major version

There's a bunch of stuff that can be further trimmed but I think I want something like style cop to do that rather than manual. Will tackle in future PR, may or may not be before next major


nunogois (approved):
LGTM!

PR #252, opened 2024-10-31
feat: allow for listing the flags known to the SDK

sighphyre:
Adds a method to list the features known to the SDK. This is intended to replace the legacy option to access the feature flags directly, which is typically used to iterate the known flags and evaluate them all

There's a bunch of noise her moving from `IsEnabled` to `Enabled` in the engine but that's all internal

Also fixes the client spec header because that's wantonly out of sync at this point

### DIFF
--- a/src/Unleash/Internal/UnleashServices.cs
+++ b/src/Unleash/Internal/UnleashServices.cs
@@ -17,7 +17,7 @@ namespace Unleash
         private readonly CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
         private readonly IUnleashScheduledTaskManager scheduledTaskManager;
 
-        public const string supportedSpecVersion = "4.5.1";
+        public const string supportedSpecVersion = "5.1.7";
 
         internal CancellationToken CancellationToken { get; }
         internal IUnleashContextProvider ContextProvider { get; }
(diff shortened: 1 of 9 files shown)


tests/Unleash.Tests/Integration/TestCaseVariant.cs
```diff
@@ -4,15 +4,9 @@ namespace Unleash.Tests.Specifications
 {
     public class TestCaseVariant
     {
-        private Variant _expectedResult;
-
         public string Description { get; set; }
         public UnleashContextDefinition Context { get; set; }
         public string ToggleName { get; set; }
-        public Variant ExpectedResult
```
sighphyre, line 12: This wasn't loading feature enabled correctly. So the spec tests haven't been checking feature_enabled correctly until now

src/Unleash/Internal/UnleashServices.cs
```diff
@@ -17,7 +17,7 @@ internal class UnleashServices : IDisposable
         private readonly CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
         private readonly IUnleashScheduledTaskManager scheduledTaskManager;
 
-        public const string supportedSpecVersion = "4.5.1";
+        public const string supportedSpecVersion = "5.1.7";
```
gastonfournier, line 20: Neat!

gastonfournier (approved):
LGTM!

PR #253, opened 2024-10-31
chore: add build that runs on mac and ubuntu too

sighphyre:
Now that we're using a native library, this becomes important

Cleans up a bunch of cruft in the build that's no longer necessary as well


gastonfournier (approved):
LG! Maybe as well bump actions/checkout to V4?

PR #254, opened 2024-10-31
chore: only send the JSON header once when sending metrics and registration

sighphyre:
Sending this multiple times breaks the Unleash header validation and metrics/registration get rejected


src/Unleash/Communication/UnleashApiClient.cs
```diff
@@ -190,7 +190,6 @@ public async Task<bool> RegisterClient(ClientRegistration registration, Cancella
             using (var request = new HttpRequestMessage(HttpMethod.Post, requestUri))
             {
                 request.Content = new StringContent(JsonSerializer.Serialize(registration, options), Encoding.UTF8, "application/json");
```
sighphyre, line 192: Needs to be set in the request content itself, otherwise it doesn't materialise as json on the receiver

daveleek (approved):
Good find! LG!

PR #255, opened 2024-10-31
chore: add dotnet core sample

sighphyre:
I used this to validate that this worked against Edge/Unleash, seems a shame to throw it away since there's no easily runnable sample in this repo at the moment


daveleek (approved):
Cool! TIL you can do await in "top-level statements" (main-less files) :)

PR #256, opened 2024-11-05
chore: swap out leaky yggdrasil custom strategy for an internal abstraction

sighphyre:
Adds an adaptor for Yggdrasil ->  Legacy strategies so we don't have to leak out Yggdrasil abstractions. The barebones interface doesn't work because the method signature requires an Yggdrasil type - the Context.

Not tested automatically because it's all just plumbing but most of the existing tests already cover the Unleash constructor and manually tested through the example project

### DIFF
--- a/src/Unleash/Internal/UnleashServices.cs
+++ b/src/Unleash/Internal/UnleashServices.cs
@@ -42,7 +42,9 @@ namespace Unleash
                 settings.FileSystem = new FileSystem(settings.Encoding);
             }
 
-            engine = new YggdrasilEngine(strategies);
+            List<Yggdrasil.IStrategy> yggdrasilStrategies = strategies?.Cast<Yggdrasil.IStrategy>().ToList();
+
+            engine = new YggdrasilEngine(yggdrasilStrategies);
 
             var backupFile = settings.GetFeatureToggleFilePath();
             var etagBackupFile = settings.GetFeatureToggleETagFilePath();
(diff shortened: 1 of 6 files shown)


src/Unleash/Internal/UnleashServices.cs
```diff
@@ -35,14 +36,16 @@ internal class UnleashServices : IDisposable
...
-        public UnleashServices(UnleashSettings settings, EventCallbackConfig eventConfig, List<IStrategy> strategies = null)
+        public UnleashServices(UnleashSettings settings, EventCallbackConfig eventConfig, List<Strategies.IStrategy> strategies = null)
         {
             if (settings.FileSystem == null)
             {
                 settings.FileSystem = new FileSystem(settings.Encoding);
             }
 
-            engine = new YggdrasilEngine(strategies);
+            List<Yggdrasil.IStrategy> yggdrasilStrategies = strategies?.Select(s => new CustomStrategyAdapter(s)).Cast<Yggdrasil.IStrategy>().ToList();
```
daveleek, line 46: the `s as IStrategy` didn't work here?

daveleek: Ah, never mind. 

> The barebones interface doesn't work because the method signature requires an Yggdrasil type - the Context.

daveleek (approved):
👍

src/Unleash/Internal/UnleashServices.cs
sighphyre: I'm not sure what you mean? Does that add anything over the cast?

PR #260, opened 2024-11-06
chore: remove broken test

sighphyre:
My bad, I was the one that suggested this test. I don't think this works correctly in a cross platform way and it's causing regular test failures on the mac-os build


daveleek (approved):
Fair enough!

PR #263, opened 2024-11-28
docs: rewrite project scoping to involve using scoped api tokens

daveleek:
Rewrites the project scoping section to revolve around using project scoped API tokens

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ The `CreateClientAsync` method was introduced in version 1.5.0, making the previ
 
 #### Project scoped Unleash client
 
-If you're organizing your feature toggles in `Projects` in Unleash Enterprise, you can [scope your API tokens](https://docs.getunleash.io/how-to/how-to-create-project-api-tokens) to just the projects needed for your client. Then use that token when configuring Unleash Client:
+If you're organizing your feature toggles in projects in Unleash Enterprise, you can [scope your API tokens](https://docs.getunleash.io/how-to/how-to-create-project-api-tokens) to include only the specific projects needed for your client. Then use that token when configuring the Unleash client:
 
 ```csharp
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -121,17 +121,20 @@ var awesome = unleash.IsEnabled("SuperAwesomeFeature");
 The `CreateClientAsync` method was introduced in version 1.5.0, making the previous `Generate` method obsolete. There's also a `CreateClient` method available if you don't prefer the async version.
 
 
-#### Configuring projects in unleash client
+#### Project scoped Unleash client
 
-If you're organizing your feature toggles in `Projects` in Unleash Enterprise, you can specify the `ProjectId` on the `UnleashSettings` to select which project to fetch feature toggles for.
+If you're organizing your feature toggles in `Projects` in Unleash Enterprise, you can [scope your API tokens](https://docs.getunleash.io/how-to/how-to-create-project-api-tokens) to just the projects needed for your client. Then use that token when configuring Unleash Client:
```
melindafekete, line 126: 1. I think generally we should use the term feature flag, instead of toggle, but since this readme is mostly using toggle, let's keep it and I'll update it in one batch at some point
2. Not sure if we need code formatting for `Projects`?
3. Can we spell client lowercase?

```suggestion
If you're organizing your feature toggles in projects in Unleash Enterprise, you can [scope your API tokens](https://docs.getunleash.io/how-to/how-to-create-project-api-tokens) to include only the specific projects needed for your client. Then use that token when configuring the Unleash client:
```

README.md
```diff
@@ -121,17 +121,20 @@ var awesome = unleash.IsEnabled("SuperAwesomeFeature");
 The `CreateClientAsync` method was introduced in version 1.5.0, making the previous `Generate` method obsolete. There's also a `CreateClient` method available if you don't prefer the async version.
 
 
-#### Configuring projects in unleash client
+#### Project scoped Unleash client
```
melindafekete, line 124:
```suggestion
#### Project-scoped Unleash client
```

melindafekete (approved):

README.md
daveleek: 1. Ugh good point, that was actually an omission on my part but it fits the overall usage of 'toggle' as you say
2. 👍 
3. 👍

PR #267, opened 2024-12-10
fix: send default buckets instead of null metrics buckets as they cause 400s

daveleek:
Initializes a metric bucket with default values if the send metrics bucket comes back as null from the engine.
This lets us keep metrics as a heartbeat even when buckets are empty

This does not introduce the 10min hold off that Ruby does. Should it?


chriswk (approved):
yeah. better than throwing errors.

PR #269, opened 2025-01-07
fix: support for utf-8 chars on win

nunogois:
https://linear.app/unleash/issue/2-3110/net-sdk-utf-8-scandinavian-character-issue

Adds support for flag names with valid UTF-8 characters like `ø` when running on Windows.

> WIP


daveleek (approved):
Nice one, good job!

PR #271, opened 2025-01-28
feat: client identification headers

Tymek:
# Description

Identification headers consistent with other Unleash SDKs.

> This PR adds standardized client identification headers to the feature and metrics calls that the client makes to Unleash. The headers are:
> - `x-unleash-appname`: the name of the application that is using the client. `UNLEASH-APPNAME` will be deleted in another PR (expand/contract pattern)
> - `x-unleash-connection-id`: an internal unique identifier for the current instance of the client generated by the built-in crypto lib
> - `x-unleash-sdk`: sdk information in the format `unleash-client-<language>:<version>`
> 
> All the headers are intended for the Unleash team. Changes should be implemented in a way that does not change SDK behavior in a significant way.
> The main use cases we have are:
> * capacity planning by knowing the number of unique connections made to the backend API
> * debugging misconfigured SDKs sending more traffic than expected

Fixes [issue/1-3275](https://linear.app/unleash/issue/1-3275/uniqueness-headers-in-net-sdk)

## Type of change

# How Has This Been Tested?

- [x] Running unit tests locally
- [ ] Starting example app and verifying headers sent over the network

**Test Configuration**:
* Linux, dotnet 6.0.135

# Checklist:

### DIFF
--- a/src/Unleash/Internal/UnleashServices.cs
+++ b/src/Unleash/Internal/UnleashServices.cs
@@ -17,6 +17,7 @@ namespace Unleash
         private static readonly ILog Logger = LogProvider.GetLogger(typeof(UnleashServices));
         private readonly CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
         private readonly IUnleashScheduledTaskManager scheduledTaskManager;
+        private readonly string connectionId = Guid.NewGuid().ToString();
 
         public const string supportedSpecVersion = "5.1.9";
 
@@ -84,6 +85,8 @@ namespace Unleash
                 {
                     AppName = settings.AppName,
                     InstanceTag = settings.InstanceTag,
+                    ConnectionId = connectionId,
+                    SdkVersion = settings.SdkVersion,
                     CustomHttpHeaders = settings.CustomHttpHeaders,
                     CustomHttpHeaderProvider = settings.UnleashCustomHttpHeaderProvider,
                     SupportedSpecVersion = supportedSpecVersion
--- a/src/Unleash/Communication/UnleashApiClient.cs
+++ b/src/Unleash/Communication/UnleashApiClient.cs
@@ -274,12 +274,20 @@ namespace Unleash.Communication
             const string userAgentHeader = "User-Agent";
             const string instanceIdHeader = "UNLEASH-INSTANCEID";
 
+            const string identifyConnectionHeader = "x-unleash-connection-id";
+            const string identifyAppNameHeader = "x-unleash-appname";
+            const string identifySdkHeader = "x-unleash-sdk";
+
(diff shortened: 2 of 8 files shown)


kwasniew (commented):
I find it weird that we can overwrite unleash-client-mock. I'd expect that it's always unleash-client-dotnet. Also how do we read the real version here?

src/Unleash/Internal/UnleashServices.cs
```diff
@@ -84,6 +85,8 @@ public UnleashServices(UnleashSettings settings, EventCallbackConfig eventConfig
                 {
                     AppName = settings.AppName,
                     InstanceTag = settings.InstanceTag,
+                    ConnectionId = connectionId,
+                    SdkVersion = settings.SdkVersion,
```
daveleek, line 89: @kwasniew don't know if this answers your questions, but the "real" header value is read here from a default implementation in UnleashSettings: https://github.com/Unleash/unleash-client-dotnet/blob/main/src/Unleash/UnleashSettings.cs#L144C31-L144C44

src/Unleash/Internal/UnleashServices.cs
```diff
@@ -17,6 +17,7 @@ internal class UnleashServices : IDisposable
         private static readonly ILog Logger = LogProvider.GetLogger(typeof(UnleashServices));
         private readonly CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
         private readonly IUnleashScheduledTaskManager scheduledTaskManager;
+        private readonly string connectionId = Guid.NewGuid().ToString();
```
daveleek, line 20: This will initiate a new connectionId per instance of the SDK which I assume is the intended behavior, ie if you create two SDK instances in the same app, these two instances will have different connection ids.

daveleek (approved):
LGTM!

src/Unleash/Internal/UnleashServices.cs
kwasniew: that's the point. each new Unleash should get unique id so that we can find wrongly scoped SDKs

Tymek: If there are 2 different "polling instances", it's correct

src/Unleash/Internal/UnleashServices.cs
kwasniew: I'd fix the original versioning to follow unleash-client-dotnet:semver convention and drop the v prefix

kwasniew: one drawback is that there's a lot of existing dotnet SDKs using the v version scheme

src/Unleash/Internal/UnleashServices.cs
daveleek: Great! Looks good!

PR #275, opened 2025-02-06
fix: prevent connection overwrite

kwasniew:
Prevent unleash-connection-id overwrite with a custom header.


daveleek (approved):
Tested locally with a quick invoke TryAdd.. multiple times and it confirms that this should work! LGTM!

PR #276, opened 2025-02-13
feat: Make the SDK follow default fetch interval

ivarconr:
# Description

Make sure the SDK follows the default fetch interval as other SDKs.


daveleek (approved):
LGTM!

PR #279, opened 2025-03-06
feat: connectionId and interval in headers

sjaanus:
Connection id is now on all headers and metrics/register endpoint.

Interval is on metrics and feature fetch.

### DIFF
--- a/tests/Unleash.Tests/Communication/UnleasRequestBodyUnitTest.cs
+++ b/tests/Unleash.Tests/Communication/UnleasRequestBodyUnitTest.cs
@@ -1,0 +1,97 @@
+﻿using System.Text.Json;
+using FluentAssertions;
+using NUnit.Framework;
+using NUnit.Framework.Internal;
+using Unleash.Communication;
+using Unleash.Metrics;
+using Unleash.Tests.Mock;
+using Yggdrasil;
+
+namespace Unleash.Tests.Communication
+{
+    public class UnleasRequestBodyUnitTest
+    {
+        private IUnleashApiClient CreateApiClient()
+        {
+            var requestHeaders = new UnleashApiClientRequestHeaders
+            {
+                ConnectionId = "00000000-0000-4000-a000-000000000000",
+            };
+
+            var httpClient = new HttpClient(messageHandler)
+            {
+                BaseAddress = new Uri("http://example.com")
+            };
+            var client = new UnleashApiClient(httpClient, requestHeaders, null);
+            return client;
+        }
+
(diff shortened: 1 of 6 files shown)


tests/Unleash.Tests/Communication/UnleasRequestBodyUnitTest.cs
```diff
@@ -0,0 +1,97 @@
...
+using NUnit.Framework;
+using NUnit.Framework.Internal;
+using Unleash.Communication;
+using Unleash.Metrics;
+using Unleash.Tests.Mock;
+using Yggdrasil;
+
+namespace Unleash.Tests.Communication
+{
+    public class UnleasRequestBodyUnitTest
```
kwasniew, line 12: this say it's a body test but the test method verifies headers

tests/Unleash.Tests/Communication/UnleasRequestBodyUnitTest.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        [SetUp]
+        public void SetupTest()
+        {
+            messageHandler = new MockHttpMessageHandler();
+        }
+
+
+        [Test]
+        public async Task IdentificationHttpHeaders()
```
kwasniew, line 84: it's difficult to tell what this test is doing. I'd expect sth like this:
* given a client with some connection id and interval
* when i send metrics or register
* we sent the connection id header and interval header

what we have instead is:
* given a generic client
* when i send metrics or register
* then registration result was boolean true

kwasniew: In other words I'd expect to see relevant test data in the test and noise hidden in helper methods. This test hides everything that is relevant (connection id and interval headers checking)

tests/Unleash.Tests/Communication/UnleasRequestBodyUnitTest.cs
```diff
@@ -0,0 +1,97 @@
...
+        {
+            api = CreateApiClient();
+
+            var engine = new YggdrasilEngine();
```
daveleek, line 88: Fair enough, saves you some setup code and shouldn't be too heavy on resources. But you could also just new up a metrics bucket here I believe

tests/Unleash.Tests/Communication/UnleasRequestBodyUnitTest.cs
sjaanus: Should be better now.

kwasniew (approved):

PR #280, opened 2025-03-06
chore(workflow): remove redundant dotnet-format tool install 

daveleek:
Removes the dotnet-format tool install step from the workflow as should be included in >= .net 6 preview 7


sighphyre (approved):
Sure, if the build is green I'm happy

PR #288, opened 2025-07-30
fix: working etag

sighphyre:
Fixes an issue where weak etags are not correctly respected by the SDK. This is an issue because Edge tends to send weak etags


sastromskis (approved):
LGTM

PR #290, opened 2025-07-30
chore: relax branch constraint on release action

sighphyre:
Should in theory let us release patches off the old v4 branch


sastromskis (approved):
LGTM

PR #292, opened 2025-08-06
fix: load backup correctly

sighphyre:


src/Unleash/Internal/CachedFilesLoader.cs
```diff
@@ -13,102 +14,138 @@ internal class CachedFilesLoader
...
+                    return jsonSerializer.Deserialize<T>(stream);
+                }
+            }
+            catch (IOException ex)
+            {
+                Logger.Warn(() => $"UNLEASH: Failed to load backup file: {path}", ex);
+                eventConfig?.RaiseError(new ErrorEvent() { ErrorType = ErrorType.FileCache, Error = ex });
+                return null;
+            }
+        }
```
nunogois, line 144: All of these seem almost generic enough to belong somewhere else, like `FileSystem.cs`. Is that worth considering?

src/Unleash/Scheduling/FetchFeatureTogglesTask.cs
```diff
@@ -100,13 +97,23 @@ public async Task ExecuteAsync(CancellationToken cancellationToken)
...
+        private string SerializeToString<T>(T obj)
+        {
+            using (var ms = new MemoryStream())
+            {
+                jsonSerializer.Serialize(ms, obj);
+                ms.Position = 0;
+                return fileSystem.Encoding.GetString(ms.ToArray());
+            }
+        }
+
```
nunogois, line 117: Will this work across different serializers?

src/Unleash/Scheduling/FetchFeatureTogglesTask.cs
```diff
@@ -81,15 +81,12 @@ public async Task ExecuteAsync(CancellationToken cancellationToken)
...
-                {
-                    jsonSerializer.Serialize(fs, result.ToggleCollection);
-                }
+                var json = SerializeToString(result.ToggleCollection);
+                fileSystem.WriteAllText(toggleFile, json);
             }
             catch (IOException ex)
             {
-                Logger.Warn(() => $"UNLEASH: Exception when writing to toggle file '{toggleFile}'.", ex);
-                eventConfig?.RaiseError(new ErrorEvent() { ErrorType = ErrorType.TogglesBackup, Error = ex });
```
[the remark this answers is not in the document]
sighphyre: Would love to, except that's a breaking change :(

PR #297, opened 2025-08-28
chore: amend uris missing trailing slash before adding relative uri

daveleek:
fixes an issue where not adding a trailing / after .../api would cause api portion to be removed before merging with relative uri when using streaming


sighphyre (approved):
LGTM

PR #299, opened 2025-09-01
fix: file loading

sighphyre:
This does a few subtle things around the way backup files are handled

1) Backup files no longer incorporate the instance id by default. This means that backups will work correctly if the user doesn't set the instance id
2) Because this is a problem for users who previously set the instance id, the code will also check for the old format is set and read from that if it is. Once the backup has been read the code will then start reading/writing to the new location
3) The SDK has no knowledge of other instances in different processes. So file contention may be a thing. We now use an atomic filesytem level write by writing to a temp file and using an atomic swap to move to the new location. This is to solve what the previous code was likely trying to solve with the instance id
4) Backup handler is now responsible for both reading and writing files. No more file system aware feature fetching tasks
5) Introduced a proper file system mock so this can be used in tests to achieve proper test isolation. A bunch of which are currently passing because they read a previous test's state from backup


tests/Unleash.Tests/Internal/ErrorEvents_Tests.cs
```diff
@@ -168,7 +170,7 @@ public void CachedFilesLoader_Raises_ErrorEvent()
         }
 
         [Test]
-        public void Engine_TakeState_InvalidJson_Throws_Raises_ErrorEvent()
```
sighphyre, line 171: This test is yeeted. There's no reason for this to exist, this is covered by Ygg itself. Probably my bad during the move to Ygg

src/Unleash/Internal/CachedFilesLoader.cs
```diff
@@ -1,113 +1,220 @@
...
 using System.IO;
-using Unleash.Events;
+using System.Linq;
 using Unleash.Logging;
-using Unleash.Scheduling;
 
 namespace Unleash.Internal
 {
-    internal class CachedFilesLoader
+    public interface BackupManager
```
daveleek, line 8: Nitpick, but we usually name interfaces prepended with an I (IBackupManager)

sighphyre: Cheers, I forgot about this madness in .NET. I'll patch it

src/Unleash/Internal/CachedFilesLoader.cs
```diff
@@ -1,113 +1,220 @@
...
+        {
+            this.settings = settings;
+            this.eventCallbackConfig = eventCallbackConfig;
+        }
+
+        public Backup Load()
+        {
+            try
             {
-                // Ensure files exists.
```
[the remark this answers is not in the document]
sighphyre: Yeah technically this is a change in behaviour but not from a user perspective I believe. I've removed it because it was doing nothing useful except taking up lines of code. Writing an empty file doesn't protect us from anything and it has the potential to wipe another process' SDK's write, which I think is a bug. The code path that handles the missing file results in the same result

PR #300, opened 2025-09-02
fix: async test for streaming is now correctly awaited

sighphyre:
The StreamingFeatureFetcher is assigned as part of an async task which is never awaited. This leads to a race condition where sometimes this object is null during test execution. This patch just forces an await to occur

Also removes another fetch test which is flaking about 80% of the time on CI and which just doesn't make a lot of sense

### DIFF
--- a/tests/Unleash.Tests/ClientFactory/SyncStartupUnitTest.cs
+++ b/tests/Unleash.Tests/ClientFactory/SyncStartupUnitTest.cs
@@ -65,17 +65,6 @@ namespace Unleash.Tests.ClientFactory
             Assert.ThrowsAsync<Exception>(async () => await unleashFactory.CreateClientAsync(settings, synchronousInitialization: true));
         }
 
-        [Test(Description = "Delayed initialization: Should only fetch toggles once")]
-        public async Task DelayedInitializationFetchCount()
-        {
-            settings.UnleashApiClient = mockApiClient;
-
-            var unleash = await unleashFactory.CreateClientAsync(settings);
-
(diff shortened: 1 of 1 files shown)


tests/Unleash.Tests/ClientFactory/SyncStartupUnitTest.cs
```diff
@@ -70,7 +70,7 @@ public async Task DelayedInitializationFetchCount()
         {
             settings.UnleashApiClient = mockApiClient;
 
-            var unleash = await unleashFactory.CreateClientAsync(settings);
+            var unleash = await unleashFactory.CreateClientAsync(settings, true);
```
sighphyre, line 73: Also a async test that wasn't being correctly awaited

tests/Unleash.Tests/ClientFactory/SyncStartupUnitTest.cs
```diff
@@ -65,17 +65,6 @@ public void ImmediateInitializationBubbleAsyncErrors()
             Assert.ThrowsAsync<Exception>(async () => await unleashFactory.CreateClientAsync(settings, synchronousInitialization: true));
         }
 
-        [Test(Description = "Delayed initialization: Should only fetch toggles once")]
```
sighphyre, line 68: This is now flaky. I don't actually see a way to make this work correctly, the test doesn't make a lot of sense and the interesting stuff is covered by other tests in this file

daveleek (approved):
One comment for you to look at. It's in a test so nothing important, but potential future flaky?

It seems by starting the StreamingFeatureFetcher from the outside StartAsync will get called twice, but I don't know what any fallout of that would look like

PR #301, opened 2025-09-02
Chore/mock out filesystem

sighphyre:
Patches the mock file system to behave more realistically, use that in some tests

There's also a behaviour change when loading bootstrap. The test says that should return an empty string when the file isn't found but there's no actual check for a missing file so presumably this crashes with an exception in live. Not actually sure how this is passing, but presumably another test is putting an empty file on the file system

### DIFF
--- a/src/Unleash/Utilities/ToggleBootstrapFileProvider.cs
+++ b/src/Unleash/Utilities/ToggleBootstrapFileProvider.cs
@@ -1,4 +1,5 @@
-﻿using Unleash.Internal;
+﻿using System.IO;
+using Unleash.Internal;
 
 namespace Unleash.Utilities
 {
@@ -15,7 +16,14 @@ namespace Unleash.Utilities
 
         public string Read()
         {
-            return settings.FileSystem.ReadAllText(filePath);
+            try
+            {
+                return settings.FileSystem.ReadAllText(filePath);
+            }
+            catch (FileNotFoundException)
+            {
+                return string.Empty;
+            }
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/Unleash/Utilities/ToggleBootstrapFileProvider.cs
```diff
@@ -15,7 +16,14 @@ internal ToggleBootstrapFileProvider(string filePath, UnleashSettings settings)
...
         {
-            return settings.FileSystem.ReadAllText(filePath);
+            try
+            {
+                return settings.FileSystem.ReadAllText(filePath);
+            }
+            catch (FileNotFoundException)
+            {
+                return string.Empty;
+            }
```
sighphyre, line 26: [Docs](https://learn.microsoft.com/en-us/dotnet/api/system.io.file.readalltext?view=net-9.0) say this raises a FileNotFoundException if this isn't found, which matches my experience of this API. I've chosen to make it work like the test says it should but I honestly have no idea how this ever passed. Pretty sure it shouldn't have

daveleek, line 26: Maybe this used to use the old mock thing under the hood that does things weirdly, and the test verified the incorrect mock 🤷

tests/Unleash.Tests/Mock/MockFileSystem.cs
```diff
@@ -1,32 +1,70 @@
...
 namespace Unleash.Tests.Mock
 {
     class MockFileSystem : IFileSystem
     {
+        internal readonly Dictionary<string, string> _fileSystem = new();
+        public Encoding Encoding => Encoding.UTF8;
+
         public bool FileExists(string path)
         {
-            return true;
```
daveleek, line 10: Be gone, and never come back

src/Unleash/Utilities/ToggleBootstrapFileProvider.cs
sighphyre: I thought about that but this seemed to call out to the real fs. If there's a swap out for the mock fs I can't spot it

PR #305, opened 2025-10-07
Fix background tasks leak when CreateClientAsync fails (issue #304)

### LINKED ISSUE #304: CreateClientAsync with synchronous Initialization: true leaves background tasks running when initialization fails
**Describe the bug**  
When calling `CreateClientAsync` with `synchronousInitialization: true`, if initialization fails, the returned client is `null`. This prevents disposing of the client and leaves background tasks running with no way to stop them.
 
**To Reproduce**  
1. Configure `UnleashSettings` with an invalid or unreachable API URL.  
2. Call the following code: 
```csharp
    var client = await unleashFactory.CreateClientAsync(settings, synchronousInitialization: true);
    
    if (client == null)
    {
        // At this point, background tasks are already running,
        // but we cannot dispose the client to stop them.
    }
```
3. Observe that client is null.
4. Background tasks remain running, but you cannot dispose of them.

**Expected behavior**
- Either always return a valid client instance, even if initialization fails, so that Dispose() can stop background tasks.
- Or, ensure background tasks are terminated if initialization fails and the client is null.

**Screenshots**
N/A

**Desktop (please complete the following information):**
- OS: Windows 11
- .NET runtime: .NET 8.0
- Package version: 5.5.0

**Smartphone (please complete the following information):**
N/A

**Additional context**
This can lead to resource leaks when initialization fails because background tasks continue running without a way to clean them up.

MohammadAminPourmoradian:
## 🧩 Scenario Explanation

For clarity, the code below has been simplified to highlight the main logic.

In our project, we intended to use the following code:

```csharp
services.AddSingleton(provider =>
{
    try
    {
        var unleashFactory = new UnleashClientFactory();
        var createTask = unleashFactory.CreateClientAsync(settings, synchronousInitialization: true);
        var unleash = createTask.GetAwaiter().GetResult();
        return unleash;
    }
    catch
    {
        DefaultUnleash unleash = new(settings,
            new NationalCodeStrategy(),
            new PhoneNumberStrategy(),
            new UserIdStrategy());
        return unleash;
    }
});
```

However, when calling:

```csharp
CreateClientAsync(settings, synchronousInitialization: true)
```

if the **Unleash server is not available**, this code throws an exception and prevents the application from starting properly.

To handle this situation gracefully, we needed a **fallback mechanism** that allows the app to use a default instance until Unleash becomes available again.

**Note:** The following error occurs inside the `catch` block when the fallback mechanism is triggered:

```
An item with the same key has already been added. Key: fetch-feature-toggles-task
```

This happened because multiple background tasks were being registered with the same key during repeated initialization attempts.

The recent change resolves this issue by properly handling background task registration and preventing duplicate additions.  
The fix has been **tested and verified** successfully.

Closes #304

### DIFF
--- a/tests/Unleash.Tests/ClientFactory/SyncStartupUnitTest.cs
+++ b/tests/Unleash.Tests/ClientFactory/SyncStartupUnitTest.cs
@@ -1,42 +1,41 @@
-﻿using FakeItEasy;
-using FluentAssertions;
-using NUnit.Framework;
+﻿using NUnit.Framework;
 using NUnit.Framework.Internal;
+using Unleash.Communication;
+using FakeItEasy;
+using Unleash.ClientFactory;
+using FluentAssertions;
+using static Unleash.Tests.Specifications.TestFactory;
 using System.Net.Http.Headers;
 using System.Text;
-using Unleash.ClientFactory;
-using Unleash.Communication;
 using Unleash.Scheduling;
-using static Unleash.Tests.Specifications.TestFactory;
 
 namespace Unleash.Tests.ClientFactory
 {
-	public class SyncStartupUnitTest
-	{
-		private IUnleashApiClient mockApiClient { get; set; }
-		private UnleashSettings settings { get; set; }
-		private IUnleashClientFactory unleashFactory { get; set; }
+    public class SyncStartupUnitTest
+    {
+        private IUnleashApiClient mockApiClient { get; set; }
+        private UnleashSettings settings { get; set; }
+        private IUnleashClientFactory unleashFactory { get; set; }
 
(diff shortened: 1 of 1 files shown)


daveleek:
Thank you for your contribution @MohammadAminPourmoradian. We'll take a proper look at it. We forgot to update the original issue, but we have also been working on a fix for the issue, just that we're taking a different approach where we clean the whole object up to avoid leaks. I'll give you an update as soon as I know more

daveleek:
We have a different fix being released for this where we clean up the object structure completely.
https://github.com/Unleash/unleash-dotnet-sdk/pull/306

PR #306, opened 2025-10-08
fix: dispose timers and fix disposal issues

daveleek:
This PR fixes the issue with setting up a new Unleash instance after failing a synchronized initialization where the timer collection would stay around after the error and prevent newing up new instances with the same unleash settings object.

This also fixes a cleanup issue where disposing timers could lead to errors

### DIFF
--- a/src/Unleash/ClientFactory/UnleashClientFactory.cs
+++ b/src/Unleash/ClientFactory/UnleashClientFactory.cs
@@ -1,4 +1,5 @@
-﻿using System.Threading;
+﻿using System;
+using System.Threading;
 using System.Threading.Tasks;
 using Unleash.Strategies;
 
@@ -25,11 +26,19 @@ namespace Unleash.ClientFactory
                 settings.ScheduleFeatureToggleFetchImmediatly = false;
                 settings.ThrowOnInitialFetchFail = true;
                 var unleash = new DefaultUnleash(settings, strategies);
-                TaskFactory
-                    .StartNew(() => unleash.services.FetchFeatureTogglesTask.ExecuteAsync(CancellationToken.None))
-                    .Unwrap()
-                    .GetAwaiter()
-                    .GetResult();
+                try
+                {
+                    TaskFactory
+                        .StartNew(() => unleash.services.FetchFeatureTogglesTask.ExecuteAsync(CancellationToken.None))
+                        .Unwrap()
+                        .GetAwaiter()
+                        .GetResult();
+                }
+                catch (Exception ex)
+                {
+                    unleash.Dispose();
+                    throw ex;
+                }
 
                 return unleash;
             }
@@ -49,7 +58,15 @@ namespace Unleash.ClientFactory
                 settings.ScheduleFeatureToggleFetchImmediatly = false;
                 settings.ThrowOnInitialFetchFail = true;
                 var unleash = new DefaultUnleash(settings, strategies);
-                await unleash.services.FetchFeatureTogglesTask.ExecuteAsync(CancellationToken.None).ConfigureAwait(false);
+                try
+                {
+                    await unleash.services.FetchFeatureTogglesTask.ExecuteAsync(CancellationToken.None).ConfigureAwait(false);
+                }
+                catch (Exception ex)
+                {
+                    unleash.Dispose();
+                    throw ex;
+                }
                 return unleash;
             }
             return new DefaultUnleash(settings, strategies);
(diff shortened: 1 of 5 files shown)


src/Unleash/ClientFactory/UnleashClientFactory.cs
```diff
@@ -25,11 +26,19 @@ public IUnleash CreateClient(UnleashSettings settings, bool synchronousInitializ
...
+                    TaskFactory
+                        .StartNew(() => unleash.services.FetchFeatureTogglesTask.ExecuteAsync(CancellationToken.None))
+                        .Unwrap()
+                        .GetAwaiter()
+                        .GetResult();
+                }
+                catch (Exception ex)
+                {
+                    unleash.Dispose();
+                    throw ex;
```
sighphyre, line 40: I think Github has a fair point here. We probably want this
```suggestion
                    throw;
```

src/Unleash/ClientFactory/UnleashClientFactory.cs
```diff
@@ -49,7 +58,15 @@ public async Task<IUnleash> CreateClientAsync(UnleashSettings settings, bool syn
...
                 var unleash = new DefaultUnleash(settings, strategies);
-                await unleash.services.FetchFeatureTogglesTask.ExecuteAsync(CancellationToken.None).ConfigureAwait(false);
+                try
+                {
+                    await unleash.services.FetchFeatureTogglesTask.ExecuteAsync(CancellationToken.None).ConfigureAwait(false);
+                }
+                catch (Exception ex)
+                {
+                    unleash.Dispose();
+                    throw ex;
```
sighphyre, line 68:
```suggestion
                    throw;
```

src/Unleash/Internal/UnleashServices.cs
```diff
@@ -174,6 +174,10 @@ public void Dispose()
             }
 
             engine?.Dispose();
+            if (scheduledTaskManager != null && !(scheduledTaskManager is SystemTimerScheduledTaskManager))
+            {
+                Logger.Warn(() => $"UNLEASH: Disposing ScheduledTaskManager of type {scheduledTaskManager.GetType().Name}");
```
sighphyre, line 179: Should this be a warn? This kind feels like info to me

src/Unleash/Scheduling/SystemTimerScheduledTaskManager.cs
```diff
@@ -80,34 +83,47 @@ async void Callback(object state)
...
         {
-            if (disposeEnded)
-                return;
+            if (_disposed) return;
+            _shuttingDown = true;
 
             var timeout = TimeSpan.FromSeconds(1);
 
-            using (var waitHandle = new ManualResetEvent(false))
+            var keys = new List<string>(timers.Keys.AsEnumerable());
```
sighphyre, line 99: Keys is a terrible name. I actually don't know what this is looking at it. Timer ids? Can we call it what it is?

src/Unleash/Scheduling/SystemTimerScheduledTaskManager.cs
```diff
@@ -80,34 +83,47 @@ async void Callback(object state)
...
                     {
-                        if (!waitHandle.WaitOne(timeout))
-                        {
-                            throw new TimeoutException($"UNLEASH: Timeout waiting for task '{task.Key}' to stop..");
-                        }
+                        if (!done.WaitOne(timeout))
+                            throw new TimeoutException($"Timeout waiting for timer '{k}' to stop.");
                     }
+
+                    timers.Remove(k);
```
sighphyre, line 120: Why this? We clear everything at the end don't we?

tests/Unleash.Tests/ClientFactory/SyncStartupUnitTest.cs
```diff
@@ -178,6 +178,30 @@ public void Synchronous_Initialization_500_Throws()
             });
         }
 
+        [Test]
+        public void Failing_Synchronous_Initialization_Can_Still_Create_New_Instance()
```
sighphyre, line 182: Noice

sighphyre (approved):
Approach looks sane. Some thoughts but up to you if you want to take them

src/Unleash/Scheduling/SystemTimerScheduledTaskManager.cs
daveleek: Mmm fair enough. I'll remove this

src/Unleash/Internal/UnleashServices.cs
daveleek: We're disposing something not created by us, that could have side effects on their side?

PR #312, opened 2025-11-12
feat: rework dotnetcore sample app to be automatable

daveleek:
Reworks the dotnet core sample app so we can distribute it as container and use it to benchmark a bit

### DIFF
--- a/samples/DotnetCore/Dockerfile
+++ b/samples/DotnetCore/Dockerfile
@@ -1,0 +1,20 @@
+FROM --platform=$BUILDPLATFORM mcr.microsoft.com/dotnet/sdk:9.0-alpine AS build
+
+ARG client_version
+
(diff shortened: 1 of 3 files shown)


samples/DotnetCore/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM --platform=$BUILDPLATFORM mcr.microsoft.com/dotnet/sdk:9.0-alpine AS build
+
+ARG client_version
```
daveleek: `docker build --tag testapp:version . --build-arg client_version=version`

ie. if you want a specific version of the .NET SDK (5.5.3 for instance)

chriswk (approved):
Looks reasonable. possibly make the version of the SDK configurable on buildtime of the image if possible?

PR #313, opened 2025-11-13
chore: make sample app run in docker as intended

daveleek:
Makes the sample app startable (.NET 9 not 8) and actually wait/keep running after being started


sighphyre (approved):
Fair enough!

PR #337, opened 2026-02-13
feat: implement impact metrics from Yggdrasil

daveleek:
Implements the collection of metrics from Yggdrasil and inserting into metric payload before sending

### DIFF
--- a/src/Unleash/Internal/UnleashConfig.cs
+++ b/src/Unleash/Internal/UnleashConfig.cs
@@ -10,6 +10,7 @@ namespace Unleash.Internal
     internal class UnleashConfig
     {
         internal string AppName { get; set; }
+        internal string Environment { get; set; }
         internal string InstanceTag { get; set; }
         internal string SdkVersion { get; set; }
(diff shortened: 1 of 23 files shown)


src/Unleash/Internal/UnleashConfig.cs
```diff
@@ -10,6 +10,7 @@ namespace Unleash.Internal
     internal class UnleashConfig
     {
         internal string AppName { get; set; }
+        internal string Environment { get; set; }
```
sighphyre: Guess who's back. Back again. Tell a friend

But seriously why _is_ this back?

sighphyre (approved):
Some details to look at here I think but nothing dangerous

sighphyre (approved):
Cheers

src/Unleash/Internal/UnleashConfig.cs
daveleek: Renamed to CurrentEnvironment and type changed to Func<string>. Implementation now also checks the custom header provider

PR #338, opened 2026-03-02
chore: post refactor cleanup, dispose scheduler properly

daveleek:
So turns out post refactor we weren't disposing the scheduler again. I've fixed that and moved the responsibility out to the DefaultUnleash instance which is the one creating the UnleashConfig so seemed apt.

I also moved the disposing of the engine to the same place.
And added the processed event id to the debug logging of SSE processing


sighphyre (approved):
Cheers

PR #341, opened 2026-04-01
feat: Support Microsoft.Extensions.Logging and add opt-in AOT support by trimming LibLog

### LINKED ISSUE #319: Fails to run in AOT compiled projects
**Describe the bug**
This has the same root cause as #148. Ultimately it's caused by using the now long deprecated LibLog.

**To Reproduce**
Include the SDK in a AOT published project and create a DefaultUnleashClient

It will crash.

**Expected behavior**
It works as expected.

**Additional context**
Replacing LibLog with the appropriate modern logging abstractions would be the best choice here I feel.

JacobMillward:
# Description

Adds opt-in Native AOT / trimming support on net8.0 without changing behaviour for existing consumers. This is a non-breaking change: LibLog stays the default, so logging keeps working with zero changes, and AOT support is opt-in.

- net8.0 logging: LibLog remains the default (reflection-based auto-detect, unchanged behaviour). A new UnleashLog dispatcher picks the backend per call - an explicit ILoggerFactory wins, else LibLog, else silent.
- `<UnleashUseLibLog>false</UnleashUseLibLog>`: opt-out MSBuild flag that trims LibLog's reflection out entirely (embedded `ILLink.Substitutions.xml` feature switch) and routes logging through Microsoft.Extensions.Logging. This is what makes AOT/trim publishing clean.
- Auto-off + warning: publishing with PublishAot/PublishTrimmed and the flag unset auto-disables LibLog and emits UNLEASH001, so it's never silent. Set the flag explicitly to silence it.
- `UnleashSettings.LoggerFactory` / `AddUnleash()`: Microsoft.Extensions.Logging support + DI auto-resolution, taking precedence over LibLog.
- Also here: net6.0 → net8.0, STJ source generators for `ClientRegistration`/`ClientMetrics`, `IsAotCompatible` on net8.0.
- netstandard2.0: untouched.

Fixes #319

> **Note:** Full AOT publish support also requires upstream changes in `Unleash.Yggdrasil` (native loader rewrite, STJ source generators). This PR addresses everything within this SDK's control. See https://github.com/Unleash/yggdrasil-bindings/pull/101

## Type of change

## How Has This Been Tested?

- Full test suite passes (332 tests, including 7 new logging tests)
- Both targeted frameworks build cleanly: `dotnet build -f netstandard2.0` and `dotnet build -f net8.0`
- New tests cover: log level mapping, exception forwarding, lazy factory resolution, `UnleashSettings.LoggerFactory` wiring, `AddUnleash()` DI integration

# Checklist:


daveleek:
Thank you for your contribution @JacobMillward! We'll have a look and get back to you!

Just a quick note that we're heading into easter, so might take a few days before you hear something

daveleek:
@JacobMillward Thank you for your work on this, we like the direction this is taking and we'd love to work with you on bringing this over the line.
But it's also something we need to allocate dedicated time for.

One of the reasons for that is that this will be a breaking change requiring a major version release, and we've had a major release less than 2 months ago.
The breaking change stems from the fact that removing Liblog and the reflection-based wiring code means logging from Unleash stops happening unless our customers make changes to their apps.
If you see a path forward without this happening we might be able to squeeze this into a minor release and could ship it faster!

We'll have another conversation internally around this early next week and will update you after that.

JacobMillward:
@daveleek After a 3 month hiatus I've come back to this!

Made some changes based on your feedback; this is no longer a breaking change, so it should be minor-releasable rather than needing a major.

- LibLog stays the default on net8.0, so existing consumers keep their logging with zero changes.
- AOT/trimming is now opt-in via a `<UnleashUseLibLog>false</UnleashUseLibLog>` MSBuild flag. That trims LibLog's reflection out entirely and routes logging through Microsoft.Extensions.Logging instead (set a `LoggerFactory`, or use `AddUnleash()`).
- When publishing with PublishAot/PublishTrimmed and the flag is unset, it auto-disables LibLog and emits a warning (new `UNLEASH001`) so it's never a silent change.
- Microsoft.Extensions.Logging support + the `AddUnleash()` DI extension are still there, they just take precedence over LibLog when configured.

Also rebased onto current main.

Hopefully this resolves the problems and provides a pathway to release? Next major release it should be relatively easy to remove LibLog if that's the desire.

jonkelling (approved):

JacobMillward:
@jonkelling assuming I fix the csproj issue, is an approval for me to merge, or the team?

daveleek:
@JacobMillward I think @jonkelling was showing support for the need.
I've just been getting back and up and running again since summer leave, and haven't had a chance to have another proper look at this yet.

I'll have another look and analysis of the impact of this, and discuss with my other SDK maintainer colleagues. Then we'll get back to you!

This is something we want to do, we just need to make sure the circumstances around who is affected and how, and blast radius are at acceptable levels before committing to it

build/Unleash.Client.targets
```diff
@@ -0,0 +1,31 @@
...
+       When publishing trimmed or Native AOT and the property is left unset, it is
+       auto-set to false (LibLog cannot survive trimming) and warning UNLEASH001 is
+       emitted so the choice is never silent. Setting the property explicitly silences
+       the warning. -->
+
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net8.0'">
+    <!-- Record whether we are about to auto-disable (unset + trimming/AOT). -->
+    <_UnleashAutoDisableLibLog Condition="'$(UnleashUseLibLog)' == '' and ('$(PublishAot)' == 'true' or '$(PublishTrimmed)' == 'true')">true</_UnleashAutoDisableLibLog>
+    <!-- Apply the default. -->
+    <UnleashUseLibLog Condition="'$(UnleashUseLibLog)' == '' and '$(_UnleashAutoDisableLibLog)' == 'true'">false</UnleashUseLibLog>
```
daveleek: Can't these two be merged into one?
```suggestion
    <UnleashUseLibLog Condition="('$(PublishAot)' == 'true' or '$(PublishTrimmed)' == 'true')">false</UnleashUseLibLog>
```

src/Unleash/Unleash.csproj
```diff
@@ -1,11 +1,15 @@
...
+       for the SDK's own build; packed into the NuGet build/ folder for consumers. -->
+  <Import Project="..\..\build\Unleash.Client.targets" />
+
   <PropertyGroup>
     <RootNamespace>Unleash</RootNamespace>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
     <IncludeSymbols>true</IncludeSymbols>
     <SymbolPackageFormat>snupkg</SymbolPackageFormat>
-    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net8.0</TargetFrameworks>
```
daveleek: Ok, so for everything to work properly we need to move to net8.0

src/Unleash/Communication/UnleashApiClient.cs
```diff
@@ -15,13 +15,16 @@
...
+using Unleash.Serialization;
+#endif
 
 namespace Unleash.Communication
 {
     internal class UnleashApiClient : IUnleashApiClient
     {
 
-        private static readonly ILog Logger = LogProvider.GetLogger(typeof(UnleashApiClient));
+        private static readonly ILog Logger = UnleashLog.GetLogger(typeof(UnleashApiClient));
```
daveleek: I notice that we stop using the LogProvider API provided by liblog.cs, but the liblog.cs file has not been cleaned up. Is this intentionally being left in place? Or could it be safely removed also?

src/Unleash/Serialization/UnleashJsonSerializerContext.cs
```diff
@@ -0,0 +1,16 @@
...
+    [JsonSerializable(typeof(ClientRegistration))]
+    [JsonSerializable(typeof(ClientMetrics))]
+    [JsonSourceGenerationOptions(
+        PropertyNamingPolicy = JsonKnownNamingPolicy.CamelCase,
+        IncludeFields = true)]
+    internal partial class UnleashJsonSerializerContext : JsonSerializerContext
+    {
+    }
+}
+#endif
```
daveleek: This seems to be an unrelated change?

PR #346, opened 2026-05-18
chore: remove some legacy examples that aren't contributing to unders…

sighphyre:
Removes two of the example projects that I don't think are helping much in terms of understanding how to use the SDK. Doing this because they light up like a super nova on some code hygiene tools I'm using and the easiest way to yank up the metrics is to yeet them


daveleek (approved):
Yes, yes we can!

PR #347, opened 2026-05-19
fix: prevent NullReferenceException in ToggleBootstrapFileProvider when using UseBootstrapFileProvider()

### LINKED ISSUE #141: Null ref in ToggleBootstrapFileProvider, FileSystem is null.
**Describe the bug**
Filesystem is internal so it cannot be created manually, and is not initialised until the Services are constructed. This means there is no way to create a ToggleBootstrapFileProvider in the UnleashSettings, and UseBootstrapFileProvider does not work because FileSystem is null. This means you cannot use the ToggleBootstrapFileProvider to generate the initial cache.

**To Reproduce**
1. Create an unleash settings, configure as normal.
2. Call UnleashSettings.UseBootstrapFileProvider
3. Create the unleash client
4. Null reference exception on ToggleBootstrapFileProvider.Read, line 32.

**Expected behavior**
Toggles should be loaded from the supplied files

**Screenshots**
image

**Additional context**
I'm using this sdk in Unity engine to connect to GitLab feature flags, which use unleash. I need to read from a local backup since devices aren't guaranteed to have internet access, and versions might be out of date, so we're taking a snapshot of all the feature flags and saving them in our ci pipeline.

I'm currently working around that problem by duplicating the ToggleBootstrapFileProvider, and using System.File to read without the IFileSystem.

PSanetra:
# Description

The `FileSystem` internal property on `UnleashSettings` was never initialized before `ToggleBootstrapFileProvider.Read()` accessed it, causing a `NullReferenceException` when using `UseBootstrapFileProvider()`.

This change makes the `FileSystem` getter lazily return a new `FileSystem(Encoding)` instance when the backing field is null, ensuring the property is always usable without requiring explicit assignment.

Fixes #141

## Type of change

# How Has This Been Tested?

- [x] Added regression test `Returns_File_Content_Without_Explicitly_Setting_FileSystem` in `ToggleBootstrapFileProvider_Tests` that verifies `UseBootstrapFileProvider()` works without explicitly setting `FileSystem`
- [x] All existing unit tests pass locally

# Checklist:

### DIFF
--- a/src/Unleash/UnleashSettings.cs
+++ b/src/Unleash/UnleashSettings.cs
@@ -105,10 +105,16 @@ namespace Unleash
         /// </summary>
         internal IUnleashApiClient UnleashApiClient { get; set; }
 
+        private IFileSystem _fileSystem;
+
         /// <summary>
         /// INTERNAL: Gets or sets the file system abstraction. Can be used for testing/mocking etc.
         /// </summary>
-        internal IFileSystem FileSystem { get; set; }
+        internal IFileSystem FileSystem
+        {
+            get => _fileSystem ?? new FileSystem(Encoding);
+            set => _fileSystem = value;
+        }
 
         /// <summary>
         /// Gets or sets the toggle bootstrap provider (file, url, etc). Can be used for testing/mocking etc.


daveleek:
Thank you @PSanetra - we'll have a look and get back to you!

daveleek (approved):
Thank you for this contribution @PSanetra! This is an area we missed during our migration to an internal config object which we'll want to use here as well further along, but right now this is good enough! I'll merge and get a release ready

PR #349, opened 2026-08-06
fix: correctly clean up temp files on failed backup

### LINKED ISSUE #348: Delete temporary cache files in case of an exception
**Describe the bug**
During the process of creating a cache file, a temporary file is created (which is later moved/renamed as the new cache file). In case of an exception, this temporary file is currently not deleted and can lead to orphan files in the temp folder.

**To Reproduce**
(we could not reproduce the error in a stable manner)

**Expected behavior**
In case of an exception/error, the temporary file is deleted

**Screenshots**
Image

**Server(please complete the following information):**
 - OS: Windows

**Suggestion**
It seems like the method 'CachedFilesLoader.WriteBackup' misses cleanup code in the catch block.

sighphyre:
Patches the backup clean up code to correctly delete the hanging temporary file rather than a non existing file when backup raises an exception 

Closes #348

### DIFF
--- a/src/Unleash/Internal/CachedFilesLoader.cs
+++ b/src/Unleash/Internal/CachedFilesLoader.cs
@@ -140,7 +140,7 @@ namespace Unleash.Internal
             }
             catch (Exception)
             {
-                try { if (fileSystem.FileExists(path)) fileSystem.Delete(path); } catch { /* swallow */ }
+                try { if (fileSystem.FileExists(tempPath)) fileSystem.Delete(tempPath); } catch { /* swallow */ }
                 throw;
             }
         }
--- a/tests/Unleash.Tests/Internal/CachedFilesLoader_Backup_And_Etag_Tests.cs
+++ b/tests/Unleash.Tests/Internal/CachedFilesLoader_Backup_And_Etag_Tests.cs
@@ -107,5 +107,39 @@ namespace Unleash.Tests.Internal
             fileSystem.ReadAllText(CachedFilesLoader.GetFeatureToggleETagFilePath(settings)).Should().Be("12345");
             fileSystem.ReadAllText(CachedFilesLoader.GetFeatureToggleFilePath(settings)).Should().Be("features");
         }
+
+        [Test]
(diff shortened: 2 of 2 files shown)


src/Unleash/Internal/CachedFilesLoader.cs
```diff
@@ -140,7 +140,7 @@ private void WriteBackup(string path, string content)
             }
             catch (Exception)
             {
-                try { if (fileSystem.FileExists(path)) fileSystem.Delete(path); } catch { /* swallow */ }
+                try { if (fileSystem.FileExists(tempPath)) fileSystem.Delete(tempPath); } catch { /* swallow */ }
```
sighphyre, line 143: Preeeeeetty sure the original author goofed here and this was always meant to be tempPath. It's the only thing that makes sense here

Since I'm the original author I asked myself and I said I agree so I guess that settles it

tests/Unleash.Tests/Internal/CachedFilesLoader_Backup_And_Etag_Tests.cs
```diff
@@ -107,5 +107,39 @@ public void Saving_Features_And_Flags_Writes_To_Underlying_FileSystem()
...
+            A.CallTo(() => fileSystem.FileOpenCreate(A<string>._))
+                .ReturnsLazily((string path) =>
+                {
+                    tempPath = path;
+                    return new MemoryStream();
+                });
+            A.CallTo(() => fileSystem.FileExists(A<string>._))
+                .ReturnsLazily((string path) => path == featureTogglePath || path == tempPath);
+            A.CallTo(() => fileSystem.Replace(A<string>._, featureTogglePath, null))
+                .Throws<IOException>();
```
sighphyre, line 133: The big mock is a big ugly. I can break it out if someone desperately wants but it seems like the simplest way right now to get this done and it's a pretty low concern test imo

gastonfournier (approved):
`tempPath` looks good to me! The original author makes a convincing case 😄

PR #350, opened 2026-08-18
chore: leave scheduler cleanup to the cleanup method

daveleek:
Fixes a scheduler issue where clearing out the scheduler while a task is running would crash the entire app.
The fix is to remove the finally clause that contains the call to stopping the timer in the dictionary, and trusting the Dispose to do the job instead

### DIFF
--- a/src/Unleash/Scheduling/SystemTimerScheduledTaskManager.cs
+++ b/src/Unleash/Scheduling/SystemTimerScheduledTaskManager.cs
@@ -24,36 +24,43 @@ namespace Unleash.Scheduling
 
             async void Callback(object state)
             {
-                if (_shuttingDown) return;
-
                 try
                 {
-                    if (!cancellationToken.IsCancellationRequested)
+                    if (_shuttingDown) return;
+
+                    try
                     {
-                        await task.ExecuteAsync(cancellationToken);
+                        if (!cancellationToken.IsCancellationRequested)
+                        {
+                            await task.ExecuteAsync(cancellationToken);
+                        }
                     }
-                }
-                catch (TaskCanceledException taskCanceledException)
-                {
-                    if (!cancellationToken.IsCancellationRequested)
+                    catch (TaskCanceledException taskCanceledException)
                     {
-                        Logger.Warn(() => $"UNLEASH: Task '{name}' cancelled ...", taskCanceledException);
+                        if (!cancellationToken.IsCancellationRequested)
+                        {
+                            Logger.Warn(() => $"UNLEASH: Task '{name}' cancelled ...", taskCanceledException);
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        Logger.Warn(() => $"UNLEASH: Unhandled exception from background task '{name}'.", ex);
+                    }
+                    finally
+                    {
+                        if (cancellationToken.IsCancellationRequested)
+                        {
+                            // Stop the timer.
+                            if (timers.TryGetValue(name, out var timerToStop))
+                            {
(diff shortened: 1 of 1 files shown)


src/Unleash/Scheduling/SystemTimerScheduledTaskManager.cs
```diff
@@ -24,37 +24,44 @@ public void ConfigureTask(IUnleashScheduledTask task, CancellationToken cancella
 
             async void Callback(object state)
             {
-                if (_shuttingDown) return;
-
                 try
                 {
-                    if (!cancellationToken.IsCancellationRequested)
+                    if (_shuttingDown) return;
```
gastonfournier: Keeping this outside the try catch should be fine, safe and maybe better.

daveleek:
There's a concern raised that this PR potentially leaks a timer.
The code removed in this PR removes a timer cancellation:

```csharp
    finally
    {
        if (cancellationToken.IsCancellationRequested)
        {
            // Stop the timer.
            if (timers.TryGetValue(name, out var timerToStop))
            {
                timerToStop.SafeTimerChange(Timeout.Infinite, Timeout.Infinite, ref _disposed);
            }
        }
    }
```

The cancellation token is used in three places in the SystemTimerScheduledTaskManager:
1. To avoid starting a new request if cancellation is requested
2. To log that cancellation of the task is requested
3. To stop the timer for this particular task (This is the part that is causing the threading issue we're fixing and removed in this PR, explanation follows)

The cancellationToken comes from DefaultUnleash' cancellationTokenSource:

```csharp
public class DefaultUnleash : IUnleash
{
// ... BuildUnleashConfig():
    return new UnleashConfig
    {
// ...
        CancellationToken = cancellationTokenSource.Token
    }
```

The cancellationTokenSource is private, and .Cancel() which is the way to request a cancellation, is called only once in the code base: 

```csharp
        public void Dispose()
        {
            if (!cancellationTokenSource.IsCancellationRequested)
            {
                cancellationTokenSource.Cancel();
            }

            services?.Dispose();
            config.Engine?.Dispose();

            if (IsCustomScheduledTaskManager)
            {
                Logger.Warn(() => $"UNLEASH: Disposing ScheduledTaskManager of type {config.ScheduledTaskManager.GetType().Name}");
            }
            config.ScheduledTaskManager.Dispose();
        }

```

This code also calls Dispose on the ScheduledTaskManager, which in turn cancels timers and clears the timer dictionary:

 ```csharp
       public void Dispose()
        {
            if (_disposed) return;

            _shuttingDown = true;
            var timeout = TimeSpan.FromSeconds(1);
            var timerNames = new List<string>(timers.Keys.AsEnumerable());

            foreach (var name in timerNames)
            {
                timers[name]?.Change(Timeout.Infinite, Timeout.Infinite);
            }

            foreach (var name in timerNames)
            {
                var t = timers[name];
                if (t is null) continue;

                using (var done = new ManualResetEvent(false))
                {
                    var willSignal = t.Dispose(done);
                    if (willSignal)
                    {
                        if (!done.WaitOne(timeout))
                            throw new TimeoutException($"Timeout waiting for timer '{name}' to stop.");
                    }
                }
            }

            timers.Clear();
            _disposed = true;
        }
```

This is also the perceived cause for the issue and this PR:
Dispose cancels the token, which if the task has been started will wait for it to complete, and then cancel the timer in the dictionary.
While this happens, Dispose moves on to cancelling all timers in the dictionary and then clears the dictionary.
If this manages to happen before the task is finished and the finally is called, the cancellation of the timer in the task will result in an attempted lookup on a dictionary that is already cleared out or in the process of being cleared out.

The PR aims to eliminate this problem by ensuring the timer only gets cleared out once.

daveleek:
Replaced by #351

PR #351, opened 2026-08-20
fix: stop scheduled timer directly on cancellation

gastonfournier:
## Context

This is an alternative exploration to #350. It tries to preserve cancellation-driven timer cleanup while avoiding callback access to the shared `timers` dictionary.

Related discussion:
- https://github.com/Unleash/unleash-dotnet-sdk/pull/350
- https://github.com/Unleash/unleash-dotnet-sdk/pull/350#issuecomment-5353437228

## Approach

- Capture the disabled `Timer` created for the callback and stop that instance directly when cancellation is observed.
- Treat `ObjectDisposedException` as the expected outcome when scheduler disposal wins the race.
- Leave `SystemTimerScheduledTaskManager.Dispose()` responsible for disposing timers and clearing the dictionary.
- Avoid a `ConcurrentDictionary` and avoid check-then-act synchronization through `_shuttingDown`.

The capture is initialized safely: the timer is created with `Timeout.Infinite`, assigned to the captured local, added to the dictionary, and only then started.

## Tests

Added focused coverage for:
- cancellation stopping a repeating timer without manager disposal;
- disposal completing while an async callback is suspended, followed by cancellation and callback continuation.

Local verification:
- `dotnet format Unleash.sln --verify-no-changes --no-restore` — passed
- Release build — passed
- focused scheduler tests — 2 passed
- full test suite under the locally installed .NET 10 runtime with major roll-forward — 401 passed, 1 unrelated streaming test failed (`Switches_To_Polling_When_Too_Many_Http_Errors`); it also fails in isolation and uses a custom scheduler rather than the changed implementation. Repository CI runs the suite on its native .NET 8 runtime.

### DIFF
--- a/src/Unleash/Scheduling/SystemTimerScheduledTaskManager.cs
+++ b/src/Unleash/Scheduling/SystemTimerScheduledTaskManager.cs
@@ -21,6 +21,7 @@ namespace Unleash.Scheduling
         public void ConfigureTask(IUnleashScheduledTask task, CancellationToken cancellationToken, bool start)
         {
             var name = task.Name;
+            Timer timer = null;
 
             async void Callback(object state)
             {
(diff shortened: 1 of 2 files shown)


daveleek (approved):
This should do the trick!

PR #352, opened 2026-08-24
chore: bump yggdrasil to version 1.3.2

daveleek:
Bumps Yggdrasil dependency to 1.3.2.


src/Unleash/Metrics/ClientMetrics.cs
```diff
@@ -25,7 +25,7 @@ public string YggdrasilVersion
         {
             get
             {
-                return "0.14.0";
+                return "0.21.5";
             }
         }
```
daveleek, line 30: This whole thing doesn't appear to be in use anywhere, I'll consider removing it

src/Unleash/Metrics/ClientRegistration.cs
```diff
@@ -31,7 +31,7 @@ public string YggdrasilVersion
         {
             get
             {
-                return "0.14.0";
+                return "0.21.5";
             }
         }
```
daveleek, line 36: This whole thing doesn't appear to be in use anywhere, I'll consider removing it

sighphyre (approved):

